PR #21, opened 2022-09-26
Remove TODO and add free-tier info

riverma:
Proposal to remove draft `TODO` text and replace with actual free-tier pricing info / link.


eaddingtonwhite (approved):
Good catch @riverma thank you so much!

PR #22, opened 2022-09-26
JS example for delete invocation typo

riverma:
I believe this is a typo within the Javascript Delete API invocation example. Should replace `get` with `delete`.


eaddingtonwhite (approved):
🦅 👁️ Thanks @riverma ! 🍻

PR #27, opened 2022-10-18
feat: break out Bandwidth and RPS limits

eaddingtonwhite:
Iterating based on feedback from customers. Mentioned having bandwidth and TPS limits combined was confusing them wanted to be broken out into separate limits.


poppoerika (approved):
Looks good to me!

PR #42, opened 2022-11-10
Fix a couple of typos

aidansteele:
I also noticed a few trailing underscores, but I didn't fix those as I wasn't sure if they were extraneous, or intended to be italics, or something else.

CleanShot 2022-11-11 at 10 16 43@2x


NoSQLKnowHow:
Thanks for fixing my mistake. I will look at the underscore as well.

NoSQLKnowHow (approved):
Yep, it's a typo

PR #43, opened 2022-11-14
Change to root directory instead of /docs

NoSQLKnowHow:
Changed the content so all content comes from the root directory instead of /docs so we can do good search verification, and there is no need for the /docs as this site will only need to be in docs-only mode.


eaddingtonwhite (dismissed):
Cool lets give this a try

eaddingtonwhite (approved):

PR #58, opened 2022-11-18
fix: add language section footer

poppoerika:
I was browsing our doc on my phone and realized that the language dropdown menu was nowhere to be found.
So I added `Language` section to the footer so that users can see the JP version of our doc.
One thing to note that is when users click one of the languages, the doc will be open in a new tab.

image

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -233,6 +233,19 @@ const config = {
               },
             ],
           },
+          {
+            title: "Language",
+            items: [
+              {
+                label: "English",
+                to: "https://docs.momentohq.com/",
+              },
+              {
+                label: "日本語",
+                to: "https://docs.momentohq.com/ja/",
+              },
+            ],
+          },
         ],
       },
       prism: {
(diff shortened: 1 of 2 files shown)


eaddingtonwhite:
@poppoerika I thing they have a special class to add it as a item in side nav on mobile
https://docusaurus.io/docs/i18n/tutorial#theme-configuration
So can add like this in docusaurus config. Maybe give this a try will be better then footer? If you think footer still looks better after trying then im fine merging this.
```typescript
module.exports = {
  themeConfig: {
    navbar: {
      items: [
        {
          type: 'localeDropdown',
          position: 'left',
        },
      ],
    },
  },
};
```

poppoerika:
@eaddingtonwhite 
I don't see any configurations for nav bar on mobile and what I found for locale dropdown configurations don't include mobile view... https://docusaurus.io/docs/api/themes/configuration#navbar-locale-dropdown

eaddingtonwhite (approved):
Seems fine for now. I worry about once we add a few more languages might get cluttered but can revisit this then.

poppoerika:
> Seems fine for now. I worry about once we add a few more languages might get cluttered but can revisit this then.

Thought the same thing. I think I found something I can experiment with: https://docusaurus.io/docs/styling-layout#mobile-view
I can try this and see how it goes.

NoSQLKnowHow (commented):
Looks good to me. merging

PR #60, opened 2022-11-19
Fix ja/intro.md by removing unnecessary tags and more

riywo:
The page contains unnecessary `<br/>` tags and some minor misses.

This commit removes `<br/>` tags and the link for guides.

### DIFF
--- a/i18n/ja/docusaurus-plugin-content-docs/current/intro.md
+++ b/i18n/ja/docusaurus-plugin-content-docs/current/intro.md
@@ -5,19 +5,23 @@ title: Momento サーバーレスキャッシュサービス
 description: Learn what you need to know to speed up your cloud application with the world's first serverless cache.
 slug: /
 ---
+
 <head>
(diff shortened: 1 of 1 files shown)


riywo:
@poppoerika Minor fixes on the existing JP page.

poppoerika (approved):
Thank you for doing this!
I think we need to add some styling to adjust where Japanese words are cut off.
At least for laptop and mobile views. Will create a ticket for that.

riywo:
It's not easy to set proper word breaks for everywhere in Japanese i.e. a non-whitespace language. I think most of technical document sites just rely on how browser treats and sometime it looks weird, of course, but I think it's OK for technical documents. It's everywhere on WWW, TBH.

poppoerika:
@riywo Thank you for the insight! Yeah, sounds good.
I think it's okay for now.

riywo:
@poppoerika Awesome! Could you merge as I don't have such permission?

PR #77, opened 2022-12-12
Fixing more broken links and clarifying meta info

NoSQLKnowHow:


static/robots.txt
```diff
@@ -1 +1,4 @@
 User-agent: *
+Sitemap: https://docs.momentohq.com/sitemap.xml
+Sitemap: https://docs.momentohq.com/ja/sitemap.xml
+Disallow:
```
eaddingtonwhite: is empty Disallow like this ok for robots.txt? Its been like 8 years since I had to deal w/ this im out of loop on best practices. Just want to double check on this w/ u.

eaddingtonwhite: Also the double Sitemap: keys dont throw it off?

eaddingtonwhite (approved):

PR #91, opened 2023-01-27
feat: add winget installation instructions for windows cli users

malandis:
We add winget as the primary installation path for Windows users. Previously the only
installation path was by downloading the installer/zip from the GitHub
release page. We demote this to an alternative path.


NoSQLKnowHow (approved):
Looks good to me!

PR #104, opened 2023-02-08
Getting started rework

pete-naylor:
Moving instruqt tracks into the page, general rework to improve flow.


docs/getting-started.md
```diff
@@ -7,22 +7,29 @@ description: Jump in and get started quickly using Momento Serverless Cache with
 ---
 
 # Getting Started with Momento Serverless Cache
-If you're looking to jump in and get started with Momento Serverless Cache with an SDK or CLI, you've come to the right place. There are multiple ways to access the service below. 
+If you're looking to jump in and get started with Momento Serverless Cache with an SDK or CLI, you've come to the right place. Want to quickly get a feel for the Momento experience without installing any software? You can do so right here [in your browser](#try-our-cli-and-an-sdk-in-your-browser). Or, if you're ready to setup your development environment and get to building your own applications with Momento you can follow the [simple instructions below](#install-the-momento-command-line-interface-cli).
 
-## Try our CLI in your browser
+## Try our CLI and an SDK in your browser
```
allenheltondev, line 12: I don't think the word "an" is necessary in the header. To me it seems to read better if it's just
```markdown
## Try our CLI and SDK in your browser
```

docs/getting-started.md
```diff
@@ -7,22 +7,29 @@ description: Jump in and get started quickly using Momento Serverless Cache with
...
 
-You can try out our CLI tool today, right in your browser.  No software installation required!
+You can try out Momento Serverless Cache today, right in your browser. No software installation required! **Note: Momento Serverless Cache is an API based serverless service. It does not deploy anything into your (AWS, GCP, Azure, etc.) account.**
```
allenheltondev: Bold emphasis seems like a bit much. I might recommend using italics (single asterisk) instead.

docs/getting-started.md
```diff
@@ -7,22 +7,29 @@ description: Jump in and get started quickly using Momento Serverless Cache with
...
 
-Head over to our [Try Momento Serverless Cache page](https://www.gomomento.com/try-momento-for-free), and
-you can experiment with Momento CLI commands in a shell session directly inside your browser.
-You can create a free auth token, create a cache, and start running `set` and `get` commands
-on your cache. It only takes a few minutes!
+First, request your free auth token, create a cache, configure your CLI, and start running `set` and `get` commands on your cache. Launch below.
```
allenheltondev, line 21: This list seems out of order to me. Shouldn't it be 
```markdown
...free auth token, configure the CLI, create a cache, and start running `set` and `get` commands...
```
Since the image you're displaying directly underneath this message has a launch button on it, I'm not sure `Launch below` is necessary

allenheltondev (changes requested):
Minor suggestions

docs/getting-started.md
NoSQLKnowHow: For the note, can we instead do the docusaruas note style? You can see an example of it here https://docs.momentohq.com/develop/api-reference/collections/dictionary#dictionaryincrement

.gitignore
```diff
@@ -19,3 +19,4 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 package-lock.json
+yarn.lock
```
NoSQLKnowHow: Please do not edit the .gitignore. Can you remove this from the commit?

docs/getting-started.md
```diff
@@ -7,22 +7,29 @@ description: Jump in and get started quickly using Momento Serverless Cache with
...
 
-**Note: Momento Serverless Cache is an API based serverless service. It does not deploy anything into your (AWS, GCP, Azure, etc.) account.**
+<a href="https://play.instruqt.com/embed/momento/tracks/sandbox-container-1challenge?token=em_14J4EucaTHzL0rXw&finish_btn_target=_top&finish_btn_text=Return+to+Docs&finish_btn_url=https%3A%2F%2Fdocs.momentohq.com%2Fgetting-started#try-our-cli-and-an-sdk-in-your-browser" target="_top"><img src="/img/cli_lab.png" alt="CLI lab" /></a>
```
cprice404, line 23: @pete-naylor thanks for this!  I think this will work better than the embed.

Does this open in a new tab or just load over the current docs tab in your browser?  If the latter, curious if you have any thoughts about doing it in a new tab vs. current tab?

cc:  @eaddingtonwhite 

thanks again for working to get these in!

docs/getting-started.md
pete-naylor: I guess I could go either way on this. But we have multiple SDKs and this demo shows one in use, right?

docs/getting-started.md
pete-naylor: Turned this into note style - looks good Kirk thanks!

docs/getting-started.md
pete-naylor: Hmm. Normally I think the order I listed is correct. In this case we're letting the CLI configure create the default cache for us. I think the order is still reasonable - I started the configure first, and the cache was created subsequent to this.

Agree the "Launch below" is not strictly necessary, but it does make things clearer and I don't want to lose anyone in the flow. You'd be amazed how people find ways to get lost! Suspect Hannah might take a pass at this and clean it all up anyway - maybe reorganize the pages and the menu as you'd suggested.

.gitignore
pete-naylor: Will add another commit to delete that line from .gitignore

docs/getting-started.md
pete-naylor: I fiddled around with some variations on this last night and this seemed to be the best approach. Gets a nice full screen view, and returns to the right place with the custom button you suggested. It opens in the same tab at _top - I like this as it keeps everything in the same flow without introducing tab forks in the road. I was concerned that somehow people may not make it back to the original getting started page if it's a separate tab. Seems like there's a good chance that this will get reworked by Hannah and others who might have better ideas - I just needed it to be functional for the moment and I am by no means an expert on this stuff!!

cprice404 (approved):

allenheltondev (approved):

PR #114, opened 2023-02-22
Removed broken nav buttons.

pete-naylor:
Some nav buttons at bottom of pages/categories were not working - cleanest and best for usability seems to be removing those superfluous links. Links now navigate forward and back within a subsection.

### DIFF
--- a/docs/develop/api-reference/index.mdx
+++ b/docs/develop/api-reference/index.mdx
@@ -1,6 +1,7 @@
 ---
 sidebar_position: 5
 title: API reference information
+pagination_prev: null
 sidebar_label: API reference
 description: Learn the API calls you need to know and how to use them with Momento Serverless Cache.
 ---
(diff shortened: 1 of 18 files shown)


cprice404:
nice!  this seems like a big step forward.  just out of curiosity, in the places where you put those `null`s would it allow us to put in a link to a specific leaf page in the next/previous section?

this is still a big improvement either way but just wondering!

NoSQLKnowHow:
Currently, I cannot get the site to build in my local install with these changes. Something to do with redirects. Investigating.

pete-naylor:
@cprice404 it will allow specific locations instead of the null - yep, just seemed like the right approach for now to page forward/back only within a section. Can change if you prefer.

cprice404:
> @cprice404 it will allow specific locations instead of the null - yep, just seemed like the right approach for now to page forward/back only within a section. Can change if you prefer.

I don't mind deferring to you folks on that decision.  I feel like I might hope/expect for it to just link to the first page in the next section but the thing I care most about is not having the misleading broken buttons there so I'm fine with whichever way y'all wanna go!  🚢 from me, except the CI is failing (I'm not familiar enough with this repo to know if that is important?)

NoSQLKnowHow (approved):
Fixed the broken links on the intro.md pages for the en and js sites and all is good to go now!

PR #155, opened 2023-04-05
chore: add us-east4 to list of GCP regions

virratanasangpunth:


NoSQLKnowHow (approved):
simple change. looks good!

PR #171, opened 2023-04-24
New getting started

NoSQLKnowHow:
Changed the getting started page to use the console and lead with the SDK instead of CLI.


docs/getting-started.md
```diff
@@ -5,209 +5,209 @@ sidebar_class_name: sidebar-item-getting-started
...
+You can try out Momento Cache today, right in your browser. No software installation required!
 
-```
-$ sleep 30 // wait for item to expire
-$ momento cache get --key test
-[2022-03-31T15:46:02Z INFO  momento::commands::cache::cache_cli] getting key: test from cache: my-first-cache
-[2022-03-31T15:46:03Z INFO  momento::commands::cache::cache_cli] cache miss
-```
+:::note
+Momento Cache is an API based serverless service. It does not deploy any resources in your (AWS, GCP, Azure, etm.) accounts.
```
eaddingtonwhite: Maybe instead of `API Based` try `Fully Managed` ? not 100% on that just API based sounds bit off to me.

eaddingtonwhite (dismissed):

docs/getting-started.md
NoSQLKnowHow: Added some text around it based on your feedback. Let me know what you think, please.

eaddingtonwhite (approved):

PR #174, opened 2023-04-25
Page for AWS Secrets Manager

NoSQLKnowHow:


docs/develop/sdks-integrations/aws-secrets-manager.md
```diff
@@ -0,0 +1,79 @@
...
+
+:::info
+
+Just as you should reuse your Momento `CacheClient` and `TopicClient` objects when possible, so should you with the Momento auth token from AWS Secrets Managers. Otherwise you risk adding cost, both in time and money, to each Momento API call for the round trip to AWS Secrets Manager.
+
+:::
+
+## Entry in AWS Secrets Manager
+
+When inserting the Momento auth token into AWS Secrets Manager, it should be as plaintext with no JSON as in this screenshot. (Token blurred for security.)
```
cprice404, line 24: cc: @rlinehan - would love your input on this.  Since the token will be able to provide expiration info etc. I just wanted to doublecheck that this is how you would imagine people storing it, vs. JSON.

docs/develop/sdks-integrations/aws-secrets-manager.md
```diff
@@ -0,0 +1,79 @@
...
+    // For a list of exceptions thrown, see
+    // https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_GetSecretValue.html
+    throw error;
+  }
+}
+
+// Call the getToken function
+async function run() {
+ const token = await getToken("Momento_Auth_Token");
+ console.log(token.SecretString);
```
cprice404: Do you want to add an example of constructing a CredProvider from this?

NoSQLKnowHow: Oh, you mean using .fromString with this?

cprice404: yessir

docs/develop/sdks-integrations/aws-secrets-manager.md
```diff
@@ -0,0 +1,79 @@
...
+sidebar_label: AWS Secrets Manager
+pagination_next: null
+title: Momento + AWS Secrets Manager
+description: Learn how to retreive your Momento Auth Token in AWS Secrets Manager.
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# Retreiving a Momento auth token from AWS Secrets Manager
```
rlinehan:
```suggestion
# Retrieving a Momento auth token from AWS Secrets Manager
```

docs/develop/sdks-integrations/aws-secrets-manager.md
rlinehan: I think it should be in JSON, or at least ones with expirations should be done with json, and if those are in json IMO we should always say it's json. This I think requires the console to return the full json though, or at least I think that's a much better user experience to do it that way.

We could have it like this for now and then when we get expiring tokens in place (should be in a week or so?) update the docs to talk about that too, and about secret rotation (should be able to link to an example lambda)?

cprice404: Well the primary use case for the format we use in the console is that users should be able to just copy/paste it directly into the SDK.  I don't think the JSON would be a better user experience there.  We could theoretically show both but I think that might just make it more confusing.

I think we need to think about the SecretsManager part in the same way.  Kirk is trying to write some example code for how to pull a secret from SecretsManager and use it for auth given today's SDKs.  This means calling `CredentialProvider.fromString(...)`.  So I think if he stores it in JSON then his example code would have to base64-encode it again before it could be used with `fromString`, right?

Perhaps you are thinking a little farther down the road when we might provide a custom CredentialProvider that knows how to consume the JSON directly, when the secret is populated by our lambda?

rlinehan: > So I think if he stores it in JSON then his example code would have to base64-encode it again before it could be used with fromString, right?

No, the JSON returned from the sdk contains the already base64 encoded version of the token. The console is currently just pulling the full token out of the JSON/object and only showing the token on the screen.

Personally I would think it would be most consistent to always show the full json and say "here's the part you include in the environment variable" but maybe we should hold off on that til we have updated the CredentialProvider to take the whole JSON?  In the CLI we decided we'd always show the full JSON.

docs/develop/sdks-integrations/aws-secrets-manager.md
NoSQLKnowHow: Added.

docs/develop/sdks-integrations/aws-secrets-manager.md
NoSQLKnowHow: Fixed.

docs/develop/sdks-integrations/aws-secrets-manager.md
cprice404: I see, thanks for the clarification!  Now I think I agree with you, but since most people will be getting their token from the console, I think for Kirk's example (which is going to be based on the console workflow and pasting something into secrets manager), we probably want to stick with the plain text path until/unless we update the console as you are suggesting.

cprice404 (approved):

docs/develop/sdks-integrations/aws-secrets-manager.md
rlinehan: I'm cool with that. I think when we had expiring tokens to the console we can revisit this because I think the console will likely need to be updated anyway.

PR #175, opened 2023-04-26
POC: docusaurus plugin that supports injecting example snippets

cprice404:
This commit shows the beginnings of how we could inject example
snippets from the SDK github repos into the docs.  For now it
just uses a hard-coded map of snippet content but this can easily
be extended to pull in the content from the appropriate files
in the various repos.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -13,6 +13,15 @@ import TabItem from '@theme/TabItem';
 
 # Getting Started with Momento Cache and Momento Topics
 
+TESTING SOME DYNAMICALLY INJECTED EXAMPLE CODE:
+
+```javascript
+%%%example:typescript:code:ListCaches%%%
+```
+
+%%%example:go:description:Logging%%% %%%example:go:code:Logging%%%
+
+
 If you're looking to jump in and get started with Momento Cache with an SDK, you've come to the right place.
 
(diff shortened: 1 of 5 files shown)


docs/getting-started.md
```diff
@@ -13,6 +13,15 @@ import TabItem from '@theme/TabItem';
...
 
+TESTING SOME DYNAMICALLY INJECTED EXAMPLE CODE:
+
+```javascript
+%%%example:typescript:code:ListCaches%%%
+```
+
+%%%example:go:description:Logging%%% %%%example:go:code:Logging%%%
+
+
```
cprice404, line 24: This is what I'm currently imagining for how we would specify the example snippets that we wanted to inject.

src/remark/inject-example-code-snippets.js
```diff
@@ -0,0 +1,34 @@
...
+}
+
+const plugin = (options) => {
+  const transformer = async (ast) => {
+    visit(ast, markdownNodeContainsExampleSnippets, (node) => {
+      console.log(`Found a markdown node that contains example snippets.  Replacing content.`);
+
+      node.value = node.value.replace(/%%%([^%]*)%%%/g, (match, exampleId) => {
+        return temporaryHardCodedCodeSnippets.get(exampleId)
+      })
```
cprice404, line 28: this regex might have to get a little fancier but something like this should suffice.

PR #176, opened 2023-04-27
feat: render "Set" examples via new SnippetResolver plugins

cprice404:
This commit does a few things:

* Adds some typescript docusaurus plugins that we can use to have
  more control over how the example code is rendered on various
  pages.
* Adds a "SnippetResolver" API which will be used to look up
  example code snippets from the SDK repos.  In this PR, it is
  only resolving a few snippets from a hard-coded data structure,
  it is not doing anything dynamic across the SDK repos.
* Adds a new JSX component called `SdkExampleSnippets`.  This is
  similar to our existing `SdkExamples` component, except that it
  (a) uses the SnippetResolver to find the example snippets, and
  (b) it does not render the tab for a given language if we do
  not have an example available.
* Replaces one instance of SdkExamples (The 'Set' API examples) with
  the new `SdkExampleSnippets` to demonstrate that the rendered
  output is the same.

### DIFF
--- a/docs/develop/api-reference/index.mdx
+++ b/docs/develop/api-reference/index.mdx
@@ -7,6 +7,7 @@ description: Learn the API calls you need to know and how to use them with Momen
 ---
 
 import { SdkExamples } from "@site/src/components/SdkExamples";
+import { SdkExampleSnippets } from "@site/src/components/SdkExampleSnippets";
 
 # API reference for Momento Cache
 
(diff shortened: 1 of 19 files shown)


docs/develop/api-reference/index.mdx
```diff
@@ -228,78 +229,9 @@ Sets the value in cache with a given Time To Live (TTL) seconds. If a value for
...
-  auth_token: MOMENTO_AUTH_TOKEN, default_ttl: DEFAULT_TTL_SECONDS
-)
-response = client.set('test-cache', 'test-key', 'test-value')
-raise response.error if response.error?
-  `}
-  cli={`momento cache set --key test-key --value test-value`}
-/>
+<SdkExampleSnippets
+  snippetId={'API_Set'}
+  />
```
cprice404, line 234: After some iteration - this is now what I'm thinking all of these tab-view examples will end up looking like in the mdx files.

eaddingtonwhite (dismissed):

malandis (approved):
Very cool

rishtigupta (approved):

eaddingtonwhite (approved):
🐐 @cprice404

PR #181, opened 2023-05-02
feat: Implement plugin and component that can look up snippets from SDK repos

cprice404:
This commit adds a plugin, inject-example-code-snippets, which does the following:

* Parses the markdown files looking for SdkExampleSnippets components
* When it finds them, it looks up the appropriate snippets in the SDK repos
* And then replaces the component with an instance of SdkExampleSnippetTabs, with code samples for each language injected.

As of this PR, the only example we're injecting is the 'Set' example, and the only repo that provides it is the JavaScript SDK.  But this should lay a foundation for migrating the remaining APIs.


eaddingtonwhite (dismissed):

plugins/example-code-snippets/src/examples/resolvers/source-parsers/languages/typescript-snippet-source-parser.ts
```diff
@@ -0,0 +1,32 @@
...
+  RegexSnippetTypeOptions,
+} from '../regex-snippet-source-parser';
+import {ExampleSnippetType} from '../../../examples';
+import * as path from 'path';
+
+export class TypescriptSnippetSourceParser extends RegexSnippetSourceParser {
+  constructor(repoSourceDir: string) {
+    const codeSnippetFiles: Array<string> = [
+      // For now we don't have any typescript examples, only javascript ones.  We
+      // can add a list of files containing TS examples in the future if we have some.
```
malandis, line 12: I don't understand this comment. The example file in the SDK repo is in TypeScript, and the JavaScript snippet parser references a `.ts` file. Hence we do have TypeScript examples and they are being called JavaScript?

plugins/example-code-snippets/src/examples/resolvers/source-parsers/regex-snippet-source-parser.ts
```diff
@@ -0,0 +1,92 @@
...
+      nextLine = contentLinesIterator.next();
+    }
+
+    return undefined;
+  }
+
+  private captureUntilEndOfSnippet(
+    remainingLines: IterableIterator<string>,
+    endRegex: RegExp,
+    numLeadingSpacesToStrip: number
```
malandis, line 78: We could infer the amount of leading whitespace to strip from the snippets themselves, ie $$min_{line \in snippet}\\{\text{amount of leading whitespace in line}\\}$$

Otherwise we should rename this to `numLeadingWhitespaceCharsToStrip` because it sounds like space symbols as opposed to any whitespace (tabs).

plugins/example-code-snippets/src/examples/resolvers/source-parsers/languages/typescript-snippet-source-parser.ts
cprice404: lol as long as the snippets don't include any type annotations, they are valid JS.  If this proves problematic I can change that file to be .js in the future but that would require monkeying with the build scripts in order to make sure that it gets run during CI so I'm just trying to do the fastest/simplest thing for now.

plugins/example-code-snippets/src/examples/resolvers/source-parsers/regex-snippet-source-parser.ts
cprice404: yeah this parameter is definitely going to change in future PRs as we add more repos in.  It will probably become a higher-order function argument so that each SDK can specify its own rules for normalizing the example lines.

eaddingtonwhite (approved):

PR #188, opened 2023-05-03
feat: generate language API support matrix

cprice404:
This commit gets rid of the hard-coded language support matrix
and adds a new page to the API reference section which includes
dynamically-generated tables indicating which languages support
each API, broken into tables for each collection type.


eaddingtonwhite (approved):
⭐

PR #197, opened 2023-05-06
feat: ja translate develop/guides/cheat-sheets/momento-cache-python-cheat-sheet

### LINKED ISSUE #192: [JP translation] Cheatsheet Python
https://github.com/momentohq/public-dev-docs/blob/main/i18n/ja/docusaurus-plugin-content-docs/current/develop/guides/cheat-sheets/momento-cache-python-cheat-sheet.md

riywo:
Close #192

### DIFF
--- a/i18n/ja/docusaurus-plugin-content-docs/current/develop/guides/cheat-sheets/momento-cache-python-cheat-sheet.md
+++ b/i18n/ja/docusaurus-plugin-content-docs/current/develop/guides/cheat-sheets/momento-cache-python-cheat-sheet.md
@@ -2,23 +2,23 @@
 sidebar_position: 1
 sidebar_label: Python
 sidebar_class_name: sidebar-item-python-cheat-sheet
-title: Cheat Sheet for Python + Momento Cache
-description: Everything to get you going with coding using Python and Momento Cache
+title: Python + Momento Cache チートシート
+description: Python で Momento Cache を使ってコーディングを始めるための全てがここにあります。
 ---
 
-# Cheat Sheet for Python with Momento Cache
+# Python + Momento Cache チートシート
 
-If you need to get going quickly with Python and Momento Cache, this page contains the basic API calls you'll need.
+このページでは、Python で Momento Cache を手早く始めてみたい時に必要となる基本的な API 呼出しを紹介しています。
 
 :::tip
 
-If you combine all of the functions on this page into one python file, you'd have a central collection of functions you can import and call from other python code. In addition, if you are using this code in production you might look to replace the print() calls with ones using the logging library (`import logging`) in python. [Click here](../../../../../../../static/code/cheat-sheets/MomentoBasics.py) to see the class file with all definitions in it.
+このページにある全ての関数を1つの python ファイルに入れれば、他の python コードからインポートして呼出しできる関数のコレクションができあがります。さらにこのコードを本番環境で使う場合には、 print() 呼出しをロギングライブラリ (`import logging`) に置き換えると良いでしょう。全ての定義を含むクラスファイルを見たければ、[こちらをクリックしてください。](../../../../../../../static/code/cheat-sheets/MomentoBasics.py)
 
 :::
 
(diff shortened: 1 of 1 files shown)


i18n/ja/docusaurus-plugin-content-docs/current/develop/guides/cheat-sheets/momento-cache-python-cheat-sheet.md
```diff
@@ -128,15 +128,15 @@ An example of incrementing a value in a key. You can pass in a positive or negat
...
 :::info
-Beyond these basic API calls check out the [API reference page](../../api-reference/index.mdx) for more information on the full assortment of Momento API calls.
+これらの基礎的な API 呼出しに加えて、Momento API 呼出しの全ての一覧がある [API リファレンスページ](../../api-reference/index.mdx)もご覧下さい。
 :::
 
-## Notes of further usage
+## さらなる利用に向けて
 
-It could be adventageous to reuse the CacheClient connection object across multiple API calls depending on how your app works. If this is something you want to do, then you should modify the client() function definition to return a CacheClient object like so:
+アプリケーションの挙動によっては、複数の API 呼出しの間で CacheClient オブジェクトを再利用した方が良い場合があるでしょう。その様な場合には、 client() 関数の定義を以下の様に変えて、CacheClient オブジェクトを返す様にすると良いでしょう:
```
riywo, line 136: BTW, I'm not sure about what this paragraph means since `client()` already looks like returning `CacheClient` object.

riywo:
@poppoerika Check this out!

i18n/ja/docusaurus-plugin-content-docs/current/develop/guides/cheat-sheets/momento-cache-python-cheat-sheet.md
poppoerika: @riywo 
[Here](https://github.com/momentohq/public-dev-docs/pull/197/files#diff-748a493d3a38089ae24cab0f541d22ffcab70a43299850aa5e4bc44ee58fac1cR30-R39), `def client()` in `MomentoBasics` class is instantiating a `CacheClient` object. 
So every time you call `MomentoCounter.client()` a new `CacheClient` object is created in these operations in the examples such as [creating a cache](https://github.com/momentohq/public-dev-docs/pull/197/files#diff-748a493d3a38089ae24cab0f541d22ffcab70a43299850aa5e4bc44ee58fac1cR48-R56) and [listing caches](https://github.com/momentohq/public-dev-docs/pull/197/files#diff-748a493d3a38089ae24cab0f541d22ffcab70a43299850aa5e4bc44ee58fac1cR64-R76).

So by changing `def client()` to return just an object `CacheClient` not an instantiated object and pass this object type to each operations I mentioned above's as their arguments, then you can pass one instantiated `CacheClient` object instead of creating a `CacheClient` object every API call.

poppoerika: But this is good feedback. 
If it doesn't make sense after reading this page, we should update it so that it's clear what we're trying to communicate.

poppoerika (approved):

i18n/ja/docusaurus-plugin-content-docs/current/develop/guides/cheat-sheets/momento-cache-python-cheat-sheet.md
riywo: You probably just need `@cache` ? https://docs.python.org/3/library/functools.html#functools.cache

poppoerika: thank you for the feedback @riywo !
It looks like the cheat sheet for the Python SDK is not clear enough to convert our intention to users.
We'll discuss internally and update this page accordingly.

PR #216, opened 2023-05-16
feat: tweak wording on limits page

eaddingtonwhite:


docs/manage/limits.md
```diff
@@ -33,4 +33,4 @@ As Momento Topics utilizes Momento Cache, all Cache limits apply to Topics as we
 
 
 ## Soft limits and support
-The limits on this page are soft limits that can be altered. If you need these limits adjusted, please reach out to [Momento Support](mailto:support@momentohq.com). Please include your customerID, the name of the cache(s) to be altered, the cloud+region the cache is located in (e.g. AWS eu-west-1), and which limits from the list you’d like increased.
\ No newline at end of file
+The limits on this page are soft limits that can be altered. If you need these limits adjusted, please reach out to [Momento Support](mailto:support@momentohq.com). Please include your email you used to sign up for Momento, the name of the cache(s) to be altered, the cloud+region the cache is located in (e.g. AWS eu-west-1), and which limits from the list you’d like increased.
```
allenheltondev: I'd rephrase to "login email" instead of "email you used to sign up for Momento" so it's less wordy. 

Also, this is probably just for my edification, users have to request increases on specific caches? From an end user perspective I would have assumed these limits are global.

eaddingtonwhite: We can do either global for all their caches or per cache. Currently depends on customer ask how we configure. As we make limits self service over time we need to think more how this is exposed.

NoSQLKnowHow (approved):
LGTM

PR #218, opened 2023-05-18
Add back CLI install info

pete-naylor:
CLI install info was accidentally dropped from the docs. Adding back info with link to the repo for details.


eaddingtonwhite (approved):
LGTM nice catch this was missing.

PR #229, opened 2023-05-24
fix: feature matrix incorrect for auth APIs, some JS APIs

cprice404:
This fixes a few issues with the feature matrix generator; it was not correctly rendering support for Auth APIs, and some of the JavaScript APIs had moved to new files.

Prior to this commit, in the language support feature matrix:

* The node.js and Web SDKs were showing that they didn't support the auth APIs, which they actually do.  This fixes that.
* I had also configured the wrong file path for topics for those two SDKs, so they were showing up in the matrix as not supporting topics when they actually do support them.
* I had also configured the wrong file path for some cache features, so they were showing up as not supporting some cache features that they actually do.

This PR fixes all the file paths for the nodejs/web SDKs so that the matrix now shows the correct APIs that they support.


bruuuuuuuce:
looks good but has some conflicts

NoSQLKnowHow (approved):

PR #234, opened 2023-05-25
chore: remove legacy snippet resolver

cprice404:
This commit removes the legacy snippet resolver.  When I first introduced the snippet resolvers, I had copy/pasted all of the existing example code that was embedded in the docs and hard-coded it into the legacy snippet resolver so that we would still have it as a fallback until we were ready to start pulling snippets from the SDK repos.

A lot of that code was out of date and broken, and we are going to start pulling code snippets from the repos within the next week.  Therefore in this commit I go ahead and remove all of the legacy snippets.  For now this means that we will only have JavaScript examples on the site, but we will start filling the other ones back in very soon.


eaddingtonwhite (dismissed):

cprice404:
rebased

eaddingtonwhite (approved):

PR #261, opened 2023-06-07
feat: ja translate introduction/common-caching-strategies

riywo:


riywo:
@poppoerika 

If translation in the images is also needed, it takes more time. Let's finish text translations first.

poppoerika (approved):

poppoerika:
@riywo 
No need to add translated images for now. Thank you!

PR #314, opened 2023-06-23
chore: Reduce the Elixir prominence in the SDK list

nand4011:
Make the Elixir SDK appear lower down in the list of SDKs in the docs. It doesn't quite deserve to be first at this time.

### DIFF
--- a/docs/develop/sdks/elixir/index.md
+++ b/docs/develop/sdks/elixir/index.md
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 1
+sidebar_position: 7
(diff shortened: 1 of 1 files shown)


docs/develop/sdks/elixir/index.md
```diff
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 1
+sidebar_position: 7
```
nand4011, line 2: This puts it down with Java and PHP.

cprice404 (approved):

PR #319, opened 2023-06-29
Fix: Removing a period from middle of sentence

FractalJackal:


FractalJackal:
Whoops I think I was supposed to make a new branch. I'll try again.

PR #320, opened 2023-06-29
Fix: Changing What to Why in ToC

FractalJackal:


FractalJackal:
Whoops I think I was supposed to create a new branch.

PR #329, opened 2023-06-30
chore: Updating to node 18 as node 16 is going EOL shortly

NoSQLKnowHow:


eaddingtonwhite:
@NoSQLKnowHow Looks good. Maybe also add to `package.json` and force version there as well so forces devs also.
```
  "engines": {
    "npm": ">=9.0.0",
    "node": ">=18.0.0"
  }
```
https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines

eaddingtonwhite (approved):

PR #407, opened 2023-08-29
chore: fix outdated constructor calls to use eager connections

cprice404:


pratik151192:
Ugh I need to reset my `Find and Replace` - or am grepping incorrectly :(
Thanks for doing this and the double-check

pratik151192 (approved):

cprice404:
> Ugh I need to reset my `Find and Replace` - or am grepping incorrectly :( Thanks for doing this and the double-check

There are way too many code snippets hard-coded into this repo, I am pretty worried that a lot of them are probably not just out of date but broken

pratik151192:
It's probably worthwhile to fund a day to replace all or most of them with the plugins

PR #415, opened 2023-09-05
chore: add docs for generateDisposableTokens

anitarua:
Updated Auth API docs for generateDisposableTokens: [#422](https://github.com/momentohq/dev-eco-issue-tracker/issues/422).

Also fixed some links to section headers, added the WriteOnly role as another managed role option, and added `endpoint` to the response object descriptions. Let me know if any of that information should be different than what's stated!

Will merge this after example code snippet is merged into JS SDK.

PR #417, opened 2023-09-06
chore: better clarify disposable token permission and scope objects

anitarua:
Follow up to #415 to better clarify DisposableTokenCachePermission vs CachePermission vs TopicPermission. They're all allowed to be put into a DisposableTokenScope object, but it wasn't very clear in the original example.


pgautier404 (approved):
LGTM!

PR #420, opened 2023-09-11
Auth updates

allenheltondev:
Restructuring a tiny bit to move auth to its own section. Also beginning to reword auth token to API key (and token) when applicable.


.vscode/settings.json
NoSQLKnowHow, line 1: Should this file really be committed to the repo?

NoSQLKnowHow (approved):

PR #437, opened 2023-09-23
chore: add auto ja translation

riywo:
This commit adds scripts and workflow to automate translation to Japanese when a PR is opened for any markdown files under `docs/`.

cc @poppoerika 

## Background
@poppoerika asked me to create a workflow to create a auto translation PR for Japanese when any markdown files are updated so that we can get notified
and catch up Japanese translation.

## Design
A new GitHub Actions workflow "Translate" will be kicked when any markdown (or MDX) files are updated by any PR against `main`.

The workflow first lists all the modified files (See `translate/list.mjs`), then call Simpleen API for each file and update translated files under `i18n/` (See `translate/translate.mjs`) . Lastly, it pushes the change to a new branch and creates another PR.

## Usage
To test these script locally:

```
> edit docs/cache/indx.md
> git add . && git commit

> node translate/list.mjs HEAD^ HEAD
docs/cache/indx.md
```

```
> export SIMPLEEN_AUTH_KEY=...
> export SIMPLEEN_TRANSLATOR_ID=...
> export TARGET_LANG=ja

> node translate/translate.mjs docs/cache/index.md
Translating docs/cache/index.md => i18n/ja/docusaurus-plugin-content-docs/current/cache/index.md

> git diff
```

Note: I'll share @poppoerika how to setup Simpleen account and GitHub Secrets separately.

## Caveats
Due to limitation of Simpleen, the auto translation PR has some manual fixes like below:

1. Frontmatter won't be translated. Thus, every PR reverts previous Frontmatter translation so that the assignees need to manually re-revert before merge.
2. MDX is not supported by Simpleen so that inline MDX code will be translated as if it is a natural language. The assignees need to re-revert here, too.
3. This workflow always runs when the original PR (branch) gets update. Thus, any new commits added to the PR will be dismissed.
4. Some markdown syntaxes doesn't work well, like `_`. The assignees need to fix them manually.

## Test
See an example PR created on our forked repository: https://github.com/opsbr/momento-public-dev-docs/pull/8

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,6 @@
 {
 	"name": "Node.js",
-	"image": "mcr.microsoft.com/devcontainers/javascript-node:18-bullseye",
+	"image": "mcr.microsoft.com/devcontainers/javascript-node:20-bullseye",
 	"features": {
 		"ghcr.io/devcontainers/features/github-cli:1": {}
 	}
--- a/.github/workflows/translate.yaml
+++ b/.github/workflows/translate.yaml
@@ -1,0 +1,44 @@
+name: Translate
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "docs/**/*.md"
+      - "docs/**/*.mdx"
+
+jobs:
+  ja:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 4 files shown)


eaddingtonwhite (approved):
Added secrets to repo. Going to merge and give this try.

PR #446, opened 2023-10-06
feat: update mvi cheat sheet for latest api

malandis:
This updates the cheat sheet for the latest API and simplifies it to
incluce standalone snippets.


malandis:
I think the Japanese translation error is expected since there is no Japanese version?

eaddingtonwhite:
hmm interesting. @poppoerika we may need to add blank page in pr to kick off initial jp translation. We can do in follow up though IMO

eaddingtonwhite (approved):

nand4011 (approved):

PR #467, opened 2023-10-10
chore: remove mvi auth page

malandis:
At this point this page is not necessary. Can re-add once the MVI auth
story is more fleshed out.


eaddingtonwhite (dismissed):

eaddingtonwhite:
@malandis there is broken link to this page still in `docs/vector-index/develop/sdks/web/index.md` we should remove there also then can merge this.

malandis:
I'm going to keep for this now as we will change our auth story shortly. The browser page cross references auth so much it doesn't make sense to remove for now.

PR #470, opened 2023-10-10
chore: generate api matrix specific to each product

anitarua:
Adds leaderboard APIs to the API matrix generator and also refactors the generator so that it can create a product-specific matrix for each product page.

Addresses #432 and in preparation for creating a new section for Leaderboard docs. Can also be used to generate a language support matrix for vector index APIs.

### DIFF
--- a/docs/cache/develop/api-reference/language-support.md
+++ b/docs/cache/develop/api-reference/language-support.md
@@ -8,4 +8,4 @@ pagination_next: null
 
 ## Current Status of API support in Momento SDKs
 
-%%%API_SUPPORT_MATRIX%%%
+%%%CACHE_API_SUPPORT_MATRIX%%%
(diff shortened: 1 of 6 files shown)


cprice404:
@anitarua this amazing, thanks so much for doing this! it was on my todo list but i was never going to get to it.

i have one minor request - can you add a new markdown page that includes the full list of APIs so that there is a spot in the docs where it's possible to see it all in one shot?

i don't know exactly where we will surface that in the docs nav, so for now you don't necessarily even have to link it anywhere, you can just put the new md file in some parent directory and make sure that we can navigate to it by URL.

that'll ensure that the generator code still works for all APIs on the same page, and we can figure out where to link it from in a subsequent PR.

anitarua:
The new page shows up at `http://localhost:3000/language_support` when I run `yarn start`

cprice404 (approved):
:ty:

PR #497, opened 2023-10-30
chore: add us-east-2 to list of available regions

pratik151192:
Noticed in multiple convos that we needed `us-east-2` updated in docs and I was in the repo so added it.


eaddingtonwhite (dismissed):

eaddingtonwhite:
I think maybe one other spot needed with host names for HTTP api as well if wanted to snag that also.

eaddingtonwhite:
https://docs.momentohq.com/cache/develop/api-reference/http-api#regions

eaddingtonwhite (approved):

PR #504, opened 2023-11-02
Updating arch icons again to include Leaderboards.

FractalJackal:


allenheltondev (approved):
i'm trusting you didn't just throw random images in here

PR #518, opened 2023-11-14
chore: update list indexes and search hit for latest api

malandis:
(1) SearchHit is a pure interface not a class, so we correct the method
names there to be properties.

(2) Update list indexes response for latest format.

### DIFF
--- a/docs/vector-index/develop/api-reference/index.md
+++ b/docs/vector-index/develop/api-reference/index.md
@@ -71,7 +71,9 @@ Lists all vector indexes.
   <summary>Method response object</summary>
 
 * Success
-    * getIndexNames(): string[]
+    * getIndexes(): VectorIndexInfo[]
+        * VectorIndexInfo:
+            * getName(): string
 * Error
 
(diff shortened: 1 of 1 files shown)


docs/vector-index/develop/api-reference/index.md
```diff
@@ -71,7 +71,9 @@ Lists all vector indexes.
   <summary>Method response object</summary>
 
 * Success
-    * getIndexNames(): string[]
+    * getIndexes(): VectorIndexInfo[]
```
eaddingtonwhite, line 74: super nit but why not `getIndexes:` instead of `getIndexes():` Just noticed had removed the `()` in below example.

malandis: `VectorIndexInfo` is a class with accessor methods while `SearchHit` is an interface. I modeled `VectorIndexInfo` after `CacheInfo` which was also a class; but `SearchHit` started as an interface 🤷 . I could add property-like accessors to `VectorIndexInfo`

eaddingtonwhite (approved):

PR #522, opened 2023-11-18
add token vending machine to unity integration

nuernber:
Adds mention of using a token vending machine in the Unity Integration docs.

Depends on https://github.com/momentohq/momento-unity-demo/pull/4 which should be merged before this PR.


cprice404 (approved):
thanks @nuernber !

PR #523, opened 2023-11-22
chore: Add operations-based limit information

rlinehan:
We're moving to rate limiting based on number of operations per second, rather than number of requests. This adds a table of all APIs and how to determine the number of operations for each API.

We will also have a graph in the console showing operations/second, but this isn't yet available (and I don't have a screenshot) so I haven't added any documentation for it yet.


docs/cache/manage/limits.md
```diff
@@ -3,24 +3,85 @@ sidebar_position: 4
...
+
+## Operations
+
+Service limits are based on the number of operations performed per second. Some cache data plane APIs can perform multiple operations in a single request. Since batch operations are more efficient, the limit cost of these APIs is discounted at a 2:1 ratio. This means every two elements will count as one towards the limiter. For example, a `SetAddElements` request adding one or two elements costs one, but with three or four elements it costs two, etc.
+
+The below table describes how the number of operations is calculated for all cache APIs.
+
+| API                      | Operations                                                      |
+| ------------------------ | ------------                                                    |
+| Set                      | 1                                                               |
```
rlinehan, line 36: Most of these are 1, so another way we could display this is only show the APIs that are not just 1 operation? But although that might look a little nicer, I think the explicitness is useful, so you can look for an API and know that it is actually 1 rather than that someone forgot to add an API to this list.

docs/cache/manage/limits.md
```diff
@@ -3,24 +3,85 @@ sidebar_position: 4
...
-| Permissions per [API key or token](./../develop/api-reference/auth.md)  (hard limit)                    | 10             |
+| Permissions per [API key or token](./../develop/api-reference/auth.md)  (hard limit)                               | 10             |
 
 ## Soft limits and support
 
 The limits on this page are soft limits that can be altered unless specifically denoted. If you need limits adjusted, please reach out to [Momento Support](mailto:support@momentohq.com). Please include your login email, the name of the cache(s) to be altered, the cloud+region the cache is located in (e.g. AWS eu-west-1), and which limits from the list you’d like increased.
+
+## Operations
+
+Service limits are based on the number of operations performed per second. Some cache data plane APIs can perform multiple operations in a single request. Since batch operations are more efficient, the limit cost of these APIs is discounted at a 2:1 ratio. This means every two elements will count as one towards the limiter. For example, a `SetAddElements` request adding one or two elements costs one, but with three or four elements it costs two, etc.
```
rlinehan: From discussions with @danielamiao I believe we want to mention the 2:1 ratio but we aren't planning on surfacing it in console or customer metrics. Not entirely sure the best way to discuss it; this is what I came up with.

docs/topics/manage/limits.md
```diff
@@ -18,7 +18,7 @@ Momento Topics seek to protect themselves and their customers when it comes to s
 
 :::info
 
-Since Topics utilizes Momento Cache, so all [Cache limits](/cache/manage/limits) apply to Topics as well.
+Since Topics utilizes Momento Cache, all [Cache limits](/cache/manage/limits) apply to Topics as well.
```
rlinehan: On line 17 above (I can't comment on it) should we update the language to `10 public operations/sec`? Not sure it matters, was just thinking for consistency.

docs/cache/manage/limits.md
danielamiao:
```suggestion
Service limits are based on the number of operations performed per second. Some cache data plane APIs can perform multiple operations in a single request. Since batch operations are more efficient, the limit cost of these APIs is discounted at a 2:1 ratio. This means every two elements will count as one operation towards the limiter. For example, a `SetAddElements` request adding one or two elements costs one operation, but with three or four elements it costs two operations, etc.
```

danielamiao (dismissed):
1 NIT, also aligned with Khawaja offline, this can be shipped!

eaddingtonwhite (approved):

PR #540, opened 2023-12-07
chore: Update matrix-generator.ts to support webhooks apis

bruuuuuuuce:


plugins/example-code-snippets/src/api-support/api-support-matrix-generator.ts
```diff
@@ -235,7 +235,15 @@ const TOPIC_API_GROUPS: Array<ApiGroup> = [
...
-    apis: ['subscribe', 'publish'],
+    apis: [
+      'subscribe',
+      'publish',
+      'listWebhooks',
+      'putWebhook',
+      'deleteWebhook',
+      'getWebhookSecret',
+      'rotateWebhookSecret',
+    ],
```
cprice404, line 246: It would be awesome to also add your `Webhooks.validateRequestSignature` thing too so that it is visible on these pages.

I know it wouldn't really fit here without some additional hackery, so it doesn't need to block this PR, but would like to know what you think about maybe trying to do something as a follow-up PR.

cprice404 (approved):
thanks

PR #559, opened 2023-12-12
fix: Fix table layouts [ja]

miura-bd:
The table was broken on the Japanese Limits page, so this is the fix


poppoerika (approved):

poppoerika:
Thank you!!

PR #602, opened 2024-01-03
chore: Clarify 2:1 ratio for operations throttling

rlinehan:
Based on feedback from @cprice404
- separate and bold the 2:1 ratio of operations/request used for throttling
- add a column to the table for whether the api is a "batch" api or not
- include that these operations are weighted at /2 in the table

### DIFF
--- a/docs/cache/manage/limits.md
+++ b/docs/cache/manage/limits.md
@@ -42,24 +42,24 @@ The below table describes how the number of operations is calculated for all cac
 | Ping                     |           | 1                                                                 |
 | ItemGetType              |           | 1                                                                 |
 | KeyExists                |           | 1                                                                 |
-| KeysExist                | X         | Number of keys in request/2                                       |
+| KeysExist                | ✅        | Number of keys in request/2                                       |
 | SetIfNotExists           |           | 1                                                                 |
 | UpdateTtl                |           | 1                                                                 |
 | IncreaseTtl              |           | 1                                                                 |
 | DecreaseTtl              |           | 1                                                                 |
 | ItemGetTtl               |           | 1                                                                 |
-| DictionaryFetch          | X         | Number of fields in response/2, or 1 if dictionary is not found   |
+| DictionaryFetch          | ✅        | Number of fields in response/2, or 1 if dictionary is not found   |
 | DictionaryGetField       |           | 1                                                                 |
-| DictionaryGetFields      | X         | Number of fields in request/2                                     |
+| DictionaryGetFields      | ✅        | Number of fields in request/2                                     |
 | DictionaryIncrement      |           | 1                                                                 |
 | DictionaryRemoveField    |           | 1                                                                 |
-| DictionaryRemoveFields   | X         | Number of fields in request/2                                     |
+| DictionaryRemoveFields   | ✅        | Number of fields in request/2                                     |
 | DictionarySetField       |           | 1                                                                 |
-| DictionarySetFields      | X         | Number of fields in request/2                                     |
+| DictionarySetFields      | ✅        | Number of fields in request/2                                     |
 | DictionaryLength         |           | 1                                                                 |
-| ListFetch                | X         | Number of elements in response/2, or 1 if list is not found       |
-| ListConcatenateBack      | X         | Number of elements in request/2                                   |
-| ListConcatenateFront     | X         | Number of elements in request/2                                   |
+| ListFetch                | ✅        | Number of elements in response/2, or 1 if list is not found       |
+| ListConcatenateBack      | ✅        | Number of elements in request/2                                   |
+| ListConcatenateFront     | ✅        | Number of elements in request/2                                   |
 | ListLength               |           | 1                                                                 |
 | ListPopBack              |           | 1                                                                 |
 | ListPopFront             |           | 1                                                                 |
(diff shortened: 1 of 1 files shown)


docs/cache/manage/limits.md
```diff
@@ -27,61 +27,63 @@ The limits on this page are soft limits that can be altered unless specifically
...
-| SortedSetFetchByScore    | Number of elements in response, or 1 if sorted set is not found |
-| SortedSetGetScore        | 1                                                               |
-| SortedSetGetScores       | Number of elements in request                                   |
-| SortedSetRemoveElement   | 1                                                               |
-| SortedSetRemoveElements  | Number of elements in request                                   |
-| SortedSetGetRank         | 1                                                               |
-| SortedSetIncrementScore  | 1                                                               |
-| SortedSetLength          | 1                                                               |
-| SortedSetLengthByScore   | 1                                                               |
+| API Name                 | Batch API | Operations                                                        |
```
rlinehan: Do we want to call this `Batch API`? At one point we were calling these "multi element APIs" rather than "batch", so should this be like `Multi Element`?

docs/cache/manage/limits.md
```diff
@@ -27,61 +27,63 @@ The limits on this page are soft limits that can be altered unless specifically
...
+| ------------------------ | ----      | ------------                                                      |
+| Set                      |           | 1                                                                 |
+| Get                      |           | 1                                                                 |
+| Delete                   |           | 1                                                                 |
+| Increment                |           | 1                                                                 |
+| Ping                     |           | 1                                                                 |
+| ItemGetType              |           | 1                                                                 |
+| KeyExists                |           | 1                                                                 |
+| KeysExist                | X         | Number of keys in request/2                                       |
```
rlinehan: This isn't 100% accurate... really it's `ceil(Number of keys in request/2)` but I'm not sure how we want to represent that in this table?

docs/cache/manage/limits.md
rlinehan: I also wonder if we're going to include the `/2` in this column if we want to call this `Weighted Operations` or something like that?

cprice404: good question. If we add `multiget` and `multiset` soon, I would think of those as "batch APIs". It's a little murkier for me whether stuff inside of a collection would also qualify as "batch". I would probably vote for multi-element but I don't feel strongly, maybe @allenheltondev and/or @danielamiao can weigh in.

docs/cache/manage/limits.md
cprice404: that's probably close enough imo

docs/cache/manage/limits.md
cprice404: i'm fine with Operations

docs/cache/manage/limits.md
```diff
@@ -27,61 +27,63 @@ The limits on this page are soft limits that can be altered unless specifically
...
+| SetIfNotExists           |           | 1                                                                 |
+| UpdateTtl                |           | 1                                                                 |
+| IncreaseTtl              |           | 1                                                                 |
+| DecreaseTtl              |           | 1                                                                 |
+| ItemGetTtl               |           | 1                                                                 |
+| DictionaryFetch          | X         | Number of fields in response/2, or 1 if dictionary is not found   |
```
cprice404: in a few places in the docs (language support matrix etc.) we are using emoji checkmarks in tables:

https://github.com/momentohq/public-dev-docs/blob/main/plugins/example-code-snippets/src/api-support/api-support-matrix-generator.ts#L520

up to you if you wanna do that.

cprice404 (dismissed):

docs/cache/manage/limits.md
rlinehan: Hm, in the context of throttling, I think `multiget` and `multiset` would behave the same way as "multi element" APIs, so maybe in this context it doesn't matter? Like, if we added those APIs I think they'd just be in this same table rather than a different table?

docs/cache/manage/limits.md
rlinehan: nice! I like that. Updated in [b92305a](https://github.com/momentohq/public-dev-docs/pull/602/commits/b92305abbe45aed100312dbda5625ffb334393ae)

docs/cache/manage/limits.md
cprice404: yes, i agree. so i think the table and the docs will only need to say one thing, either way. just a matter of deciding which one of the two things to say, in case the word "batch" has other connotations/meanings to some people like it kinda does for me.

but i don't feel strongly about it.

cprice404 (dismissed):

docs/cache/manage/limits.md
danielamiao: I also prefer multi-element because something like `DictionaryFetch` doesn't intuitively sound like a "batched" API to me.

rlinehan: Renamed to multi-element in [fe61606](https://github.com/momentohq/public-dev-docs/pull/602/commits/fe61606206dd47c9f27c4197f2eaac13b5bba999)

docs/cache/manage/limits.md
```diff
@@ -27,61 +27,63 @@ The limits on this page are soft limits that can be altered unless specifically
 
 ## Operations
 
-Service limits are based on the number of operations performed per second. Some cache data plane APIs can perform multiple operations in a single request. Since batch operations are more efficient, the limit cost of these APIs is discounted at a 2:1 ratio. This means every two elements will count as one operation towards the limiter. For example, a `SetAddElements` request adding one or two elements costs one operation, but with three or four elements it costs two operations, etc.
+Service limits are based on the number of operations performed per second. Some cache data plane APIs can perform multiple operations in a single request.
+
+Since batch operations are more efficient, the limit cost of these APIs is discounted at a **2:1 ratio**. This means every two elements will count as one operation towards the limiter. For example, a `SetAddElements` request adding one or two elements costs one operation, but with three or four elements it costs two operations, etc.
```
danielamiao: NIT: `batch operations` => `multi-element operations`

danielamiao (dismissed):

docs/cache/manage/limits.md
rlinehan:
```suggestion
Since multi-element operations are more efficient, the limit cost of these APIs is discounted at a **2:1 ratio**. This means every two elements will count as one operation towards the limiter. For example, a `SetAddElements` request adding one or two elements costs one operation, but with three or four elements it costs two operations, etc.
```

rlinehan: done in [916a011](https://github.com/momentohq/public-dev-docs/pull/602/commits/916a011d858e56cabf2cc76b75f79865707123ef)

danielamiao (approved):

PR #604, opened 2024-01-05
updating the webhooks page to include additional data about payloads …

allenheltondev:
…and headers

Added a new component to see if we like it. It's a reference card to give some more dimension and professionalism to our docs. 

I also added some additional SEO tags to this page that we need to start using across the board.


docs/topics/webhooks/index.mdx
```diff
@@ -0,0 +1,133 @@
...
+
+The name of the cache the topic belongs to. This is used for namespacing purposes in Momento Topics. All topics must be published to a valid cache in your account.
+
+#### topic
+
+The name of the topic that triggered the webhook. Remember, topic names are dynamic and are not resources in your Momento account. You can publish a message to any string, which adds an incredible amount of flexibility to messages, like using a variable as the topic name of your workflow.
+
+#### event_timestamp
+
+Timestamp in milliseconds since the Unix Epoch of when the event was published.
```
FractalJackal: To match the following, should remove "of": "...since the Unix Epoch when the event..."

FractalJackal (commented):
Added one comment on a small fix to line 65.

FractalJackal (dismissed):

docs/topics/webhooks/index.mdx
```diff
@@ -0,0 +1,133 @@
...
+hide_title: true
+sidebar_position: 3
+sidebar_label: Webhooks
+sidebar_class_name: sidebar-item-webhooks-overview
+pagination_prev: null
+keywords:
+  - topics
+  - webhooks
+  - eda
+  - momento
```
cprice404, line 14: are these the SEO tags?

allenheltondev: Yes @cprice404 , those are SEO tags. They get added to the meta info for the page

allenheltondev:
The ReferenceCard was added for dimensionality in our docs. Here's what it will look like when rendered:
image

and on hover

image

docs/topics/webhooks/index.mdx
cprice404: cool. so we should be keeping a closer eye on these as PRs come in going forward?

Any guidance on how to know what things to put in here? And do they have to be single words? I would have guessed we might want the longer-form "Event Driven Architecture" here?

cprice404 (dismissed):
lgtm. I'd still like to align on best practices for these keywords going forward, to make sure folks on my team are doing the right things, but we don't need to hold up this PR for that!

woo

docs/topics/webhooks/index.mdx
allenheltondev: I'll work with @FractalJackal and @hannahjak on this, @cprice404. We can put something together and in a central location. I'll also add `event-driven architecture` as a keyword. Good callout

cprice404 (approved):

PR #614, opened 2024-01-11
chore: document score threshold in mvi search

malandis:
Adds documentation for the `scoreThreshold` MVI search parameter.


cprice404 (approved):
score

PR #668, opened 2024-01-19
chore: Move sdks to top level

cprice404:
The thrust of this PR is to remove the "SDKs" sections from under the "Develop" menu for each product; this brings us back to a world where we only have one "SDKs" page rather than a copy for every language for every product. So this would unify them back into a single page per SDK language rather than separate ones for Cache/Topic/etc.

The nav to reach the SDKs page is now in the top right, next to "Blog", etc. This is what @allenheltondev and I saw for other docs sites that we admire.

There is a new splash page for SDKs, which contains button links for relevant things for each SDK.


cprice404:
partial screenshot of what the current SDK landing page looks like. (don't laugh!)

image

src/components/LinkButton/LinkButton.module.css
```diff
@@ -0,0 +1,7 @@
+.link_button {
```
cprice404: will need help with styling here

docs/sdks/index.md
```diff
@@ -0,0 +1,191 @@
+---
```
cprice404, line 1: this is the page that I need the most input/feedback on.

allenheltondev:
I updated this PR with button styling and the link to the SDKs in top nav
image

allenheltondev (approved):

PR #684, opened 2024-01-23
Update footer text color and size

allenheltondev:
Making the theming more consistent and on brand


allenheltondev:
Footer looks like this now
image

eaddingtonwhite (approved):

PR #693, opened 2024-01-30
chore: move patterns to root and rename async processing

pratik151192:
Screenshot 2024-01-30 at 1 27 51 PM


allenheltondev (changes requested):
Please add a redirect for the asynchronous processing page to the new location. That will be in the `docusaurus.config.js` file at the root

pratik151192:
@allenheltondev added.. i had added it but forgot to commit :|

allenheltondev (approved):

PR #694, opened 2024-01-31
feat: add guide for token-id, webhooks and topics

pratik151192:
- feat: add guide for token-id, webhooks and topics


docs/topics/patterns/token_id_webhook.md
```diff
@@ -0,0 +1,367 @@
+---
+sidebar_position: 6
+title: Leverage Momento Topics, Webhooks, and Disposable Tokens in Event-Driven Architectures
```
allenheltondev: We don't follow normal title case. Can you make this
`Leverage Momento Topics, webhooks, and tokens in your event-driven architectures`

docs/topics/patterns/token_id_webhook.md
```diff
@@ -0,0 +1,367 @@
...
+description: Learn best practices around using Momento Topics, webhooks, and disposable tokens to process unique resources in your event-driven architectures.
+keywords:
+ - topics
+ - webhooks
+ - eda
+ - momento
+ - event-driven architecture
+---
+
+## Introduction to Momento Topics and Webhooks
```
allenheltondev: This should be an h1, meaning only one `#` symbol.

Also, we don't need to intro it here because we've already done that in the overview page of the docs. Instead, set the stage for what this pattern is and why you would use it.

docs/topics/patterns/token_id_webhook.md
```diff
@@ -0,0 +1,367 @@
+---
+sidebar_position: 6
```
allenheltondev, line 2: This whole thing reads like a blog to me, which is not quite what we're looking for in docs anymore. This would be a great post for something like dev.to, but on the docs site we need to be much more matter of fact, data-driven, and less storytelling.

Can you give this pattern from Twilio a read and draw some inspiration from it? https://www.twilio.com/docs/messaging/guides/preventing-messaging-fraud

Happy to help restructure if you need it, just DM me in slack.

allenheltondev (changes requested):

docs/topics/patterns/token_id_webhook.md
pratik151192: sure thing! i'll read that and structure it similarly

docs/topics/patterns/token_id_webhook.md
```diff
@@ -0,0 +1,361 @@
...
+ - topics
+ - webhooks
+ - eda
+ - momento
+ - event-driven architecture
+---
+
+# Leverage Momento Topics, webhooks, and disposable tokens in your event-driven architectures
+
+Leveraging Momento Topics, webhooks, and disposable tokens enables secure and efficient data processing within event-driven systems. With Momento Topics, you can subscribe to messages on a topic and publish messages to a new topic. Webhooks serve as HTTP callbacks triggered in response to messages published to these topics, acting as stateless consumers. Disposable Tokens are designed to provide short-lived session tokens for users and services interacting with a system. By embedding unique identifiers (token_id) in each token, they ensure secure and traceable access to resources.
```
allenheltondev: This first sentence seems to be a duplicate of the title, i'd remove id.

Also, we are just calling tokens "tokens" instead of "disposable tokens", so we need to update the verbiage there.

allenheltondev: And please link the the page on tokens when you mention them

docs/topics/patterns/token_id_webhook.md
```diff
@@ -0,0 +1,361 @@
...
+
+The key here is that you can use Topics to facilitate real-time communication between users by granting them short-lived disposable tokens. These tokens can be embedded with user information (token_id), which, when messages are published, can be leveraged to access shared resources like Momento caches via webhook callbacks. For instance, you can personalize user experiences by accessing their information stored in a Momento cache by identifying the user through the token_id. Embedding information provides two significant advantages:
+- It enhances security and prevents a user from spoofing their identity.
+- It reduces data transfer costs as the user information is embedded with the token itself.
+
+![Architecture](@site/static/img/topics/patterns/token-id-webhook.png)
+
+## Pre-requisites
+1. A public facing endpoint to receive webhook events. This endpoint must accept POST requests and be able to receive inbound calls from Momento. More detail about the structure of this event is [described here](https://docs.momentohq.com/topics/webhooks/overview#example-event).
```
allenheltondev: The link can be a local reference, which i'd prefer to do in case we ever change the docs site for any reason

docs/topics/patterns/token_id_webhook.md
```diff
@@ -0,0 +1,361 @@
...
+      configuration: Configurations.Lambda.latest(),
+      credentialProvider: CredentialProvider.fromString({apiKey: momentoApiKey}),
+      defaultTtlSeconds: 60,
+    });
+  }
+  return _cacheClient;
+}
+```
+
+5. Finally, below has sample code for generating the disposable token and publishing user messages on a Topic. In a real-world,
```
allenheltondev: This also says disposable token. In case I missed it somewhere else, please update all references to just say "token"

allenheltondev: In this sentence "topic" should be lower case since it's not referring to the product

docs/topics/patterns/token_id_webhook.md
```diff
@@ -0,0 +1,361 @@
...
+```
+### Conclusion
+
+By integrating Momento Topics, webhooks, and disposable tokens, you can create secure and stateless asynchronous systems. This convention can be applied to a variety of use-cases such as:
+
+- Multi-language [chat application](https://www.gomomento.com/blog/how-to-use-webhooks-and-momento-topics-to-build-a-multi-language-chat-app)
+- Live commenting system
+- Online polls
+
+`token-id` can be used to securely identify users sending messages in all the above use-cases.
```
allenheltondev: This makes me think we should make `token-id` formatted as code throughout the entire page

allenheltondev (changes requested):

docs/topics/patterns/token_id_webhook.md
pratik151192: I did a `grep` and can confirm it's formatted as such

docs/topics/patterns/running-background-tasks.md
```diff
@@ -14,7 +14,7 @@ keywords:
 
 # Use Momento Topics to process data asynchronously
 
-With Momento Topics, you can subscribe to messages on a topic, as well as publish messages to a new topic. Webhooks allow you to connect these topics to stateless consumers, which can then process these events asynchronously. Whether this is aggregating events by `topic_id`, saving each event to a database, or using the payload to trigger a Step function, webhooks give you the flexibility to process these events how you want to. 
+With Momento [Topics](../index.md), you can subscribe to messages on a topic, as well as publish messages to a new topic. [Webhooks](../webhooks/overview.mdx) allow you to connect these topics to stateless consumers, which can then process these events asynchronously. Whether this is aggregating events by `topic_id`, saving each event to a database, or using the payload to trigger a Step function, webhooks give you the flexibility to process these events how you want to.
```
allenheltondev: Please make the link text Momento Topics instead of just Topics.

Also update "a Step function" to "a Step Function workflow"

docs/topics/patterns/token_id_webhook.md
```diff
@@ -0,0 +1,361 @@
...
+ - topics
+ - webhooks
+ - eda
+ - momento
+ - event-driven architecture
+---
+
+# Leverage Momento Topics, webhooks, and tokens in your event-driven architectures
+
+With Momento [Topics](../index.md), you can subscribe to messages on a topic and publish messages to a new topic. [Webhooks](../webhooks/overview.mdx) serve as HTTP callbacks triggered in response to messages published to these topics, acting as stateless consumers. [Tokens](../../cache/develop/authentication/tokens.md) are designed to provide short-lived session tokens for users and services interacting with a system. By embedding unique identifiers (`token_id`) in each token, they ensure secure and traceable access to resources.
```
allenheltondev: Make this link text Momento Topics instead of just topics. I also don't really like the way the first sentence is phrased. Instead of "to a new topic" maybe make it say "to a different topic".

docs/topics/patterns/token_id_webhook.md
```diff
@@ -0,0 +1,361 @@
...
+
+The key here is that you can use Topics to facilitate real-time communication between users by granting them short-lived tokens. These tokens can be embedded with user information (`token_id`), which, when messages are published, can be leveraged to access shared resources like Momento caches via webhook callbacks. For instance, you can personalize user experiences by accessing their information stored in a Momento cache by identifying the user through the `token_id`. Embedding information provides two significant advantages:
+- It enhances security and prevents a user from spoofing their identity.
+- It reduces data transfer costs as the user information is embedded with the token itself.
+
+![Architecture](@site/static/img/topics/patterns/token-id-webhook.png)
+
+## Pre-requisites
+1. A public facing endpoint to receive webhook events. This endpoint must accept POST requests and be able to receive inbound calls from Momento. More detail about the structure of this event is [described here](../webhooks/overview#example-event).
```
allenheltondev: public-facing should be hyphenated

docs/topics/patterns/token_id_webhook.md
```diff
@@ -0,0 +1,361 @@
...
+
+        new CfnOutput(this, 'FunctionUrl ', { value: lambdaUrl.url });
+    }
+}
+
+```
+
+4. Add code to the webhook to process incoming messages. Below is sample code for the webhook lambda handler that extracts
+a user's `token_id` from the webhook payload, and access resources stored in Momento cache. It also verifies that the webhook caller
+is indeed Momento through the signing secret.
```
allenheltondev: Should this be on a new line?

docs/topics/patterns/token_id_webhook.md
```diff
@@ -0,0 +1,361 @@
...
+      _cachedSecrets.set(secretName, secretResponse.SecretString!);
+    } else {
+      throw new Error(`Unable to retrieve secret: ${secretName}`);
+    }
+  }
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  return _cachedSecrets.get(secretName)!;
+}
+
+function verifySignature(signature: string, secret: string, body: string) {
```
allenheltondev: Let's use the built-in verify instead of re-implementing it ourselves.

docs/topics/patterns/token_id_webhook.md
```diff
@@ -0,0 +1,361 @@
...
+      console.log('Successfully persisted comment for course');
+    } else if (listResp instanceof CacheListPushFront.Error) {
+      console.log('Error while publishing comment for course ' + listResp.message());
+    }
+
+    return {
+      statusCode: 200,
+      headers: {
+        'Content-Type': 'application/json',
+        'Access-Control-Allow-Origin': '*',
```
allenheltondev: You don't need the `Access-Control-Allow-Origin` header because this endpoint isn't being called from a browser.

docs/topics/patterns/token_id_webhook.md
```diff
@@ -0,0 +1,361 @@
...
+    } else {
+      throw new Error(`Unable to retrieve secret: ${secretName}`);
+    }
+  }
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  return _cachedSecrets.get(secretName)!;
+}
+
+```
+### Conclusion
```
allenheltondev: Since this isn't a blog post, I wouldn't use `Conclusion` here. Instead, maybe say `Key Takeaways` or some other wording to highlight the key points

allenheltondev: On second thought, maybe we boil it down to a couple of small bullet points then add a `Keep Reading` section that uses the ReferenceCard widget to navigate people to other content, like the multi language chat app, security post, etc... [Example](https://github.com/momentohq/public-dev-docs/blob/main/docs/topics/webhooks/overview.mdx#L110)

allenheltondev (changes requested):

allenheltondev (approved):

PR #698, opened 2024-02-06
Updating instances of RPS to OPS

FractalJackal:
Since we limit on operations, not requests (there can be multiple operations in a request) we're auditing to make sure we're consistently speaking in terms of operations and OPS. I also caught a spelling error.

### DIFF
--- a/docs/cache/manage/pricing.md
+++ b/docs/cache/manage/pricing.md
@@ -34,7 +34,7 @@ If you need further assistance (or just can’t believe it), drop a message in o
 <summary>Can I run a production app for free on Momento Cache and Topics?</summary>
 Absolutely! Our free tier and low usage tiers are just billing. It is the same exact service and features whether you use 40GB/month or 40TB/month. You get all our availability features like multi-AZ replication, hot key protection, and automatic scaling to handle bursts. In the free tier, you even get all our security features for free, such as <a href="./../develop/authentication">fine-grained auth tokens</a>, end-to-end encryption, per-request authentication, and TLS.
 
-We frequently find customers provisioning an entire cluster to handle low operations per second (OPS) workflows. If you want them to be HA, you need multiple nodes. If you believe in CICD, you likely have a similarly sized cluster in staging—and maybe even in dev. That adds up! Turn off those machines—and save the environment. Go Momento!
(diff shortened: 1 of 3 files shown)


docs/cache/manage/pricing.md
```diff
@@ -34,7 +34,7 @@ If you need further assistance (or just can’t believe it), drop a message in o
...
 Absolutely! Our free tier and low usage tiers are just billing. It is the same exact service and features whether you use 40GB/month or 40TB/month. You get all our availability features like multi-AZ replication, hot key protection, and automatic scaling to handle bursts. In the free tier, you even get all our security features for free, such as <a href="./../develop/authentication">fine-grained auth tokens</a>, end-to-end encryption, per-request authentication, and TLS.
 
-We frequently find customers provisioning an entire cluster to handle low-RPS workflows. If you want them to be HA, you need multiple nodes. If you believe in CICD, you likely have a similarly sized cluster in staging—and maybe even in dev. That adds up! Turn off those machines—and save the environment. Go Momento!
+We frequently find customers provisioning an entire cluster to handle low operations per second (OPS) workflows. If you want them to be HA, you need multiple nodes. If you believe in CICD, you likely have a similarly sized cluster in staging—and maybe even in dev. That adds up! Turn off those machines—and save the environment. Go Momento!
```
eaddingtonwhite: nit but I think this should be like `operations per second (ops/s)`

docs/cache/manage/pricing.md
```diff
@@ -52,7 +52,7 @@ Heck yeah! You are billed for the inbound and outbound data transfer, not the vo
 
 <details>
   <summary>Will it scale?</summary>
-Hell yes! Momento services are the best way to future-proof your application story. You can add Momento Cache *extremely* quickly, and it just fades into the background whether you are doing 1 RPS or 1 million RPS. You just pay for what you use, so you get a scalable cache without spending a lot of cash.
+Hell yes! Momento services are the best way to future-proof your application story. You can add Momento Cache *extremely* quickly, and it just fades into the background whether you are doing 1 OPS or 1 million OPS. You just pay for what you use, so you get a scalable cache without spending a lot of cash.
```
eaddingtonwhite: same here I think OPS is bit confusing vs TPS or RPS would just just `ops/s` to anotate

FractalJackal:
I updated the branch w/ ops/s (and I also returned "low-operations per second workflows" to be "low-request per second (RPS) workflows" because based on definitions from Tony, RPS is still a good way to characterize workloads between client and server. The ops/s one in Pricing is important to change since it's related to cost/tier and we limit on ops/s.

Not sure if I need to do a new PR for that or how to make it show that the branch is updated so we can merge it?

eaddingtonwhite (approved):

PR #704, opened 2024-02-09
chore: add style guide for node.js

cprice404:
This commit adds a new page, "Style Guide", for the node.js SDK.
It explains the tradeoffs between our traditional "pattern matching"
code style, vs. a more traditional "simplified" code style.


rishtigupta (dismissed):

allenheltondev (approved):
I like this. I think we should make a video format of this to explain the concepts and embed it in the docs. But for now, it looks good.

PR #705, opened 2024-02-12
chore: mvi rename get/delete item parameter from ids to filter

malandis:
To standardize how we select items in get item batch, get item
metadata batch, and delete item batch, we use the term filter in
methods, whether selecting by ids or filter expression.


bruuuuuuuce (approved):
:shipit:

PR #709, opened 2024-02-15
chore: convert markdown to html before displaying

rishtigupta:
## PR Description:
Convert the markdown response from the "https://robomo-ls.mochat.momentohq.com/rag-momento-vector-index/invoke" API to html before displaying.

## Reference:
Notice how the response is now being displayed using lists and paragraphs:

Screenshot 2024-02-14 at 5 19 48 PM


anitarua (approved):
🎉

PR #736, opened 2024-02-20
fix: add redirect for php cheat sheet indexed by google

anitarua:
Adds a redirect for the PHP cheat sheet page that was indexed by google search


bruuuuuuuce (approved):
:shipit:

malandis (approved):

PR #758, opened 2024-03-07
chore: add setBatch and getBatch to cache API reference

anitarua:


docs/cache/develop/api-reference/index.mdx
```diff
@@ -86,6 +86,16 @@ Sets the value in cache with a given Time To Live (TTL) seconds. If a value for
...
 <SdkExampleTabs snippetId={'API_Set'} />
 
+### SetBatch
+
+Sets multiple key-value pairs in a cache with a given Time To Live (TTL) seconds. If a value for a key is already present it will be replaced by the new value.
+
+| Name       | Type   | Description                                   |
+| ---------- | ------ | --------------------------------------------- |
+| cacheName  | String | Name of the cache.                            |
+| items      | { String/[]Byte :  String/[]Byte } | The mapping of keys to values that should be stored |
```
anitarua, line 96: not sure how best to represent this type (in the SDK it's a [Record or Map](https://github.com/momentohq/client-sdk-javascript/blob/2090e9b3d1b1e59c5a6627f807ef5c755b408b5c/packages/core/src/internal/clients/cache/AbstractCacheClient.ts#L252-L254))

closest example similar to this API I've found is [dictionarySetFields](https://docs.momentohq.com/cache/develop/api-reference/dictionary-collections#dictionarysetfields) which also accepts [Map or Record in the SDK](https://github.com/momentohq/client-sdk-javascript/blob/2090e9b3d1b1e59c5a6627f807ef5c755b408b5c/packages/core/src/internal/clients/cache/AbstractCacheClient.ts#L797-L800) but not listed as such in the reference

bruuuuuuuce: hmmmmm, these are a couple options that come to mind, I am not in love with any of them tho

```
{ String/Byte[] : String/Byte[] }
Record<String | Byte[], String | Byte[]>
Map<String | Byte[], String | Byte[]>
```

The first option seems to be the most consistent with how we do it in other places

docs/cache/develop/api-reference/index.mdx
```diff
@@ -97,6 +118,27 @@ Get the cache value stored for the given key.
...
+| --------- | ------ | --------------------------------------------- |
+| cacheName | String | Name of the cache.                            |
+| keys       | String[] / Bytes[] | The list of keys for which to retrieve values. |
+
+<details>
+
+<summary>Method response object</summary>
+
+* Success
+- `values()`: Record<string, string>
```
pgautier404:
```suggestion
- `values()`: Record&lt;string, string&gt;
```
It looks like there may be an issue here with the `<` and `>` characters in what is technically an XML snippet.

pgautier404 (approved):

PR #472, opened 2023-10-11
chore: add new docs section for Leaderboard

anitarua:
Addresses [#468](https://github.com/momentohq/dev-eco-issue-tracker/issues/468). Adds a new section for Leaderboard and an initial set of API reference pages.

PR #760, opened 2024-03-11
chore: remove leaderboards develop folder

anitarua:
Restructuring to remove the "Develop" folder.
Also restructures the Japanese docs because the build failed locally otherwise.


allenheltondev (dismissed):
Make sure you add redirects for all of these as well. That will be in the `docusaurus.confnig.js` file at the root of the repo.

I don't think you can do wildcards, but might be worth a shot?

anitarua:
@allenheltondev I don't see any redirect link for leaderboards. I think the leaderboard docs were added after that revision. 
* #431 in Sept 2023 splits docs into separate products, introducing the redirects I think
* #472 in Oct 2023 adds leaderboard docs

eaddingtonwhite (approved):

PR #820, opened 2024-03-25
feat: add green theme as background to docs

pratik151192:
Added green background to docs. The RGB for the background gradient has been taken from @FractalJackal to make sure its consistent with the gradient across other websites. 

Screenshot 2024-03-25 at 7 31 21 PM
Screenshot 2024-03-25 at 7 31 29 PM


rishtigupta:
If we make this change, we need to ensure all pages are readable. Example, the following page looks rather odd since it makes it difficult to read the sdk icons:

Screenshot 2024-03-25 at 3 37 54 PM

pratik151192:
@rishtigupta thanks for checking the icons page! I did check code and tables etc looked okay didnt notice this one. Will take a look.

rishtigupta:
I feel this change might have messed up the margins for "Ask Mo":
Screenshot 2024-03-25 at 3 44 54 PM

The production website looks to have the correct margins:
Screenshot 2024-03-25 at 3 45 46 PM

cprice404:
> If we make this change, we need to ensure all pages are readable. Example, the following page looks rather odd since it makes it difficult to read the sdk icons:
> Screenshot 2024-03-25 at 3 37 54 PM

FWIW I think these icons are already kinda messed up in dark mode:

image

 we should fix them but idk if we need to block this PR on it.

cprice404 (dismissed):
i'm good with this when @allenheltondev and @rishtigupta are. thanks for doing it

src/css/custom.scss
```diff
@@ -105,6 +105,45 @@ h4 {
...
+}
+
+/* Adjusting the navbar to blend with the background */
+/* Adjusting the navbar to blend with the new background gradient */
+/* Adjusting the navbar to blend with the background */
+.navbar {
+  background-image: linear-gradient(50deg, #0E2515, #25392b);
+  backdrop-filter: blur(5px); /* If you want a blur effect */
+  z-index: 100;
+  /* Other styles */
```
rishtigupta: Can we remove these comments from the css file?

rishtigupta:
This change would enforce dark mode for our users. I think a fast follow-up PR would be nice to fix the sdk icons. 
I am good with the pr if the margins for "Ask Mo" are corrected. Thanks!

pratik151192:
@rishtigupta @cprice404 I have already gotten the white SVGs from Hannah so I'll update them. I am not seeing the errors with margin on my end so not sure why you are seeing Rishti but will check

src/css/custom.scss
pratik151192: ok

rishtigupta (approved):

PR #824, opened 2024-03-26
fix: add missing control plane methods to cache API matrix and use correct cache client file for nodejs sdk parsing

anitarua:
Noticed missing API matrix entries for `createCache`, `deleteCache`, and `listCaches`.
Also fixed a discrepancy in client files crawled for NodeJS and Web SDKs.

Screenshot 2024-03-26 at 4 31 16 PM


cprice404 (approved):
thank you!
yes

PR #835, opened 2024-04-01
Add a pattern doc on how to do api caching

virratanasangpunth:


docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,165 @@
+---
+sidebar_position: 10
+title: Read-aside API Caching
+sidebar_label: Read-aside API Caching
+description: Learn best practices around using Momento Cache to cache results from an external API
+---
+
+# Use Momento Cache to cache results from an API
+
+Remote calls to an API can be slow and expensive. Depending on the access pattern, it can be desirable to cache results in a cache to improve your application's latency and availability.
```
allenheltondev: I'd say "store results in a cache" instead of "cache results in a cache"

docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,165 @@
...
+
+With [Momento Cache](../), you can easily cache results from an API in a highly-available and scalable remote cache.
+
+## Getting Started
+
+In this example we will build an application that can give us a string description of the current weather of the given place. We will use WolframAlpha API in this example to show you how we can cache the results of an API.
```
allenheltondev: I'd add verbiage in here explaining that the API is slow to return (give an estimate) and that since it's weather, the results are highly cacheable because it doesn't change regularly.

docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,165 @@
...
+    if (response.ok) {
+      return await response.text();
+    } else {
+      const errorString = await response.text();
+      throw new Error(`Wolfram returned an error: ${errorString}`);
+    }
+  }
+}
+```
+4. Since the weather is unlikely to change that frequently for a particular city, and since a call to a remote API can be expensive, we can cache the results using [Momento Javascript SDK](../../sdks).
```
allenheltondev: Reword this slightly to say something like, "remember, weather does not change regularly...." because we should have stated this in the intro paragraph of the section.

docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,165 @@
...
+For code reusability, we use the Decorator Pattern to wrap `CachingWeatherDescriptionClient` around an abstract `WeatherDescriptionClient` that we have created in an earlier step. Note that this requires no code-change to the existing `WolframWeatherDescriptionClient`.
```
allenheltondev: I'd put a link to some external docs defining the decorator pattern

docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,165 @@
...
+
+  async getWeatherDescription(city: City): Promise<string> {
+    const getResponse = await this.momentoCacheClient.get(
+      CACHE_NAME,
+      this.cacheKey(city)
+    );
+
+    if (getResponse instanceof CacheGet.Hit) {
+      // If the weather is in the Cache, return the result from the Cache.
+      return getResponse.valueString();
```
allenheltondev: Use getResponse.value() here instead, please

docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,165 @@
...
+  implements WeatherDescriptionClient
+{
+  delegate: WeatherDescriptionClient;
+  momentoCacheClient: CacheClient;
+  constructor(
+    delegate: WeatherDescriptionClient,
+    momentoCacheClient: CacheClient
+  ) {
+    this.delegate = delegate;
+    this.momentoCacheClient = momentoCacheClient;
```
allenheltondev: I recommend using this instead since we're pushing that as the best practice in JS/TS.
```
this.momentoCacheClient = momentoCacheClient.cache(CACHE_NAME);
```
Then all references to it no longer have to pass in the cache name.

allenheltondev: Full disclosure, i don't remember if `.cache` is an async call or not. We need to verify.

docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,165 @@
...
+    );
+
+    if (getResponse instanceof CacheGet.Hit) {
+      // If the weather is in the Cache, return the result from the Cache.
+      return getResponse.valueString();
+    } else if (getResponse instanceof CacheGet.Miss) {
+      // If the weather is not in the Cache, call actual client,
+      // and populate the Cache.
+      const response = await this.delegate.getWeatherDescription(city);
+      await this.populateCacheWithResult(city, response);
```
allenheltondev: Since we aren't using the response from this call, we likely don't need to use `await` here. Curious on your thought there.

docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,165 @@
...
+}
+```
+5. By using the Decorator Pattern, we can easily compose `CachingWeatherDescriptionClient` and `WolframWeatherDescriptionClient` together to create a new client that does read-side caching, i.e. if a result is the cache, return the result, otherwise fallback to retrieving the result from the remote API.
+This is an example code that leverages Momento Cache to cache results of WolframAlpha API into a remote cache:
+```typescript
+import {CacheClient, Configurations, CredentialProvider} from '@gomomento/sdk';
+
+async function main() {
+  const momentoClient = await CacheClient.create({
+    configuration: Configurations.Laptop.v1(),
```
allenheltondev: Initializing the client is much simpler now. You can do it like this:
```
const momentoClient = await CacheClient.create({
  credentialProvider: CredentialProvider.fromEnvVar('MOMENTO_API_KEY'),
  defaultTtlSeconds: WEATHER_CACHE_TTL_SECONDS
});
```

docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,165 @@
...
+  // "The weather in Phoenix, Arizona, currently includes no precipitation with clear skies, a wind speed of 9 miles per hour and a temperature of 78 degrees Fahrenheit"
+  console.log(`${response}`);
+}
+
+```
+
+That is it! This is a simple example of how to cache an API result. Here's a few more things that you can tweak further to adapt Momento Cache to your exact use-case:
+1. You can further leverage the Decorator pattern to add more functionality without modifying the existing code. For example you can add a Decorator for metrics/logging so that you get instrumentation on error rates and cache hit rates.
+2. TTL (how long items get to live in a cache) can be adjusted on a per-item basis to fit your use-case. For example, in our earlier code, we could've made the TTLs different for different cities.
+3. You can cache more than just Strings into Momento Cache. In our earlier example we cache strings but Momento Cache can also accept byte arrays. This allows you to cache an arbitrary object in your application, provided that you implement your own custom serialization/deserialization.
```
allenheltondev: You should add a section down here similar to the *See More* section in the Topics token_id_webhook.md page for additional reading. I recommend these:

* title: 6 6ommon caching strategies
  * link: https://www.gomomento.com/blog/6-common-caching-design-patterns-to-execute-your-caching-strategy
  * description: Learn more about other types of caching strategies
*  title:  Using Momento to cache chatbot calls
  * link: https://www.gomomento.com/resources/case-studies/cydas-people-takes-momento-cache-to-prod-for-their-chatgpt-ai-powered-chatbot-in-just-two-hours
  * description: See how CYDAS used Momento to cache calls to LangChain to build a chatbot

docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,165 @@
+---
+sidebar_position: 10
+title: Read-aside API Caching
+sidebar_label: Read-aside API Caching
+description: Learn best practices around using Momento Cache to cache results from an external API
```
allenheltondev: This needs a `keywords` property with the following tags
```yaml
keywords:
  - cache
  - momento
  - read-aside
  - wolframalpha
  - api caching
  - api

File: docs/cache/pattern/api-caching.md [hunk 11]
```diff
@@ -0,0 +1,165 @@
+---
+sidebar_position: 10
+title: Read-aside API Caching
```
allenheltondev: Update the metadata to the following:
```yaml
title: Caching API calls to WolframAlpha using Momento Cache
sidebar_label: Caching WolframAlpa API calls
description: Learn how to use Momento Cache to boost performance of applications using the WolframAlpha API
```

allenheltondev (changes requested):

docs/cache/pattern/api-caching.md
virratanasangpunth: https://github.com/momentohq/public-dev-docs/pull/835/files#diff-203fb71916cea1e9bf73c40ec3268ea35304f910d22501715e31079a54c7bce2R24

docs/cache/pattern/api-caching.md
virratanasangpunth: Hmmm yea that is a good callout. We should not `await` here and let the cache be populated _eventually_.

docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,193 @@
+---
+title: Caching API calls to WolframAlpha using Momento Cache
+sidebar_label: Caching WolframAlpa API calls
```
allenheltondev: WolframAlpha is spelled wrong here

docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,193 @@
...
+
+Remote calls to an API can be slow and expensive. Depending on the access pattern, it can be desirable to store results in a cache to improve your application's latency and availability.
+
+With [Momento Cache](../), you can easily cache results from an API in a highly-available and scalable remote cache.
+
+## Getting Started
+
+In this example we will build an application that can give us a string description of the current weather of the given place. We will use WolframAlpha API in this example to show you how we can cache the results of an API.
+
+Weather data is relatively static over short periods, making it an ideal candidate for caching. Additionally WolframAlpha API can be slow to respond due to its computational model. You can expect latency of around 700ms for WolframAlpha API versus the latency of less than 5ms for Momento GET API.
```
allenheltondev: It feels like you're missing the word "the" in a couple spots:
* "the WolframAlpha API can be slow..."
* "less than 5ms for the Momento GET API"

allenheltondev: Same on line 28, "to call the WolframAlpha API"

docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,193 @@
...
+    if (response.ok) {
+      return await response.text();
+    } else {
+      const errorString = await response.text();
+      throw new Error(`Wolfram returned an error: ${errorString}`);
+    }
+  }
+}
+```
+4. Recall that this scenario is well-suited for caching. We can cache the results using [Momento Javascript SDK](../../sdks).
```
allenheltondev: JavaScript has a capital S

docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,193 @@
...
+<div style={{
+  display: 'flex',
+  flexDirection: 'row',
+  flexWrap: 'wrap',
+  gap: '20px',
+  marginBottom: '20px' // Add margin bottom to the container if needed
+}}>
+
+  <ReferenceCard
+    title="6 common caching strategies "
```
allenheltondev: There's an extra space at the end of this title

allenheltondev (changes requested):

docs/cache/pattern/api-caching.md
virratanasangpunth: TIL lol

docs/cache/pattern/api-caching.md
virratanasangpunth: oopsie.

docs/cache/pattern/api-caching.md
```diff
@@ -0,0 +1,193 @@
...
+      const response = await this.delegate.getWeatherDescription(city);
+      this.populateCacheWithResult(city, response);
+      return response;
+    }
+
+    return weatherDescription;
+  }
+}
+```
+5. By using the Decorator Pattern, we can easily compose `CachingWeatherDescriptionClient` and `WolframWeatherDescriptionClient` together to create a new client that does read-aside caching, i.e. if a result is the cache, return the result, otherwise fallback to retrieving the result from the remote API.
```
allenheltondev: I missed this the first time, you're missing a word "if a result is **in** the cache"

virratanasangpunth: ooof, good catch, ty

allenheltondev (approved):
Great work! This looks awesome :heart:

PR #849, opened 2024-04-03
fix: add redirects from old SDK links to new ones

cprice404:
A while back we moved the SDK pages up to the top level, from
the `develop` directory. This commit adds redirects so that any
lingering references to the old links won't be broken.


cprice404:
this is the first of 2 PRs to fix https://github.com/momentohq/client-sdk-python/pull/449

virratanasangpunth (approved):

PR #898, opened 2024-04-22
Instant messaging pattern

allenheltondev:
This will be used as a reference for all other patterns going forward. This level in the taxonomy is meant to provide an overview with minimal specific implementation details.

Beneath this page will live multiple "how-to" articles that are code-based and SEO driven.


docs/topics/patterns/instant-messaging.mdx
```diff
@@ -0,0 +1,355 @@
...
+
+### Message storage
+
+Most instant messaging platforms store messages to allow users to view the conversation history if messages were sent prior to their arrival. Part of the startup tasks when entering a chat room is loading said history and displaying it on screen. This can easily be done with Momento in two ways: with a **sorted set** or a **list** cache item, each with their own tradeoffs.
+
+#### Sorted sets
+
+Typically used for things like [leaderboards](https://www.gomomento.com/blog/build-on-momento-tips-and-tricks-for-a-lightning-fast-game-leaderboard), a [sorted set](../../cache/develop/basics/datatypes#sorted-sets) is a collection data type that stores multiple value/score elements. In each element, the value is a `string` and the score is an `integer`. Specifically with instant messaging, you can store the time the message was sent as the *score* and a stringified JSON object as the *value*. Storing the message timestamp as the score will automatically allow you to sort messages by time since sorted sets do that by default.
```
pratik151192: nit: `score` is technically a floating point, but we can say in this context integers would do.

docs/topics/patterns/instant-messaging.mdx
```diff
@@ -0,0 +1,355 @@
...
+
+A [list](../../cache/develop/basics/datatypes#lists) is a collection data type that stores an ordered array of elements. You can push elements to either the front or back of a list and you can add multiple entries at one time. A list is a great data type for instant messaging because it stores messages in the order they were sent. When users join an instant messaging session, you can quickly fetch the entire chat history with a single [listFetch API call](../../cache/develop/api-reference/list-collections#listfetch).
+
+Elements in a list do not need to be unique. You can push the same message into a list over and over again and it will be added every time. Lists are a good choice for instant messaging for the following use cases:
+
+* Paging messages N at a time
+* Adding multiple messages at once
+* Maintaining a chat history of fixed length, i.e. the last 100 messages
+
+When using a list, you have more array manipulation functions than with a sorted set, allowing you to automatically truncate message history if it gets too long or remove specific elements for things like content moderation.
```
pratik151192: Ideally if we had server side API akin to `ZREMRANGEBYSCORE `, we could do this with sorted sets too.

docs/topics/patterns/instant-messaging.mdx
```diff
@@ -0,0 +1,355 @@
...
+
+A significant point to remember - **messages must be unique in sorted sets**! This means when saving the message to the sorted set, you should include some sort of message identifier or timestamp in your JSON.
+
+#### List
+
+A [list](../../cache/develop/basics/datatypes#lists) is a collection data type that stores an ordered array of elements. You can push elements to either the front or back of a list and you can add multiple entries at one time. A list is a great data type for instant messaging because it stores messages in the order they were sent. When users join an instant messaging session, you can quickly fetch the entire chat history with a single [listFetch API call](../../cache/develop/api-reference/list-collections#listfetch).
+
+Elements in a list do not need to be unique. You can push the same message into a list over and over again and it will be added every time. Lists are a good choice for instant messaging for the following use cases:
+
+* Paging messages N at a time
```
pratik151192: nit: Can we say `Sending N messages` here?

docs/topics/patterns/instant-messaging.mdx
```diff
@@ -0,0 +1,355 @@
...
+
+Most instant messaging platforms store messages to allow users to view the conversation history if messages were sent prior to their arrival. Part of the startup tasks when entering a chat room is loading said history and displaying it on screen. This can easily be done with Momento in two ways: with a **sorted set** or a **list** cache item, each with their own tradeoffs.
+
+#### Sorted sets
+
+Typically used for things like [leaderboards](https://www.gomomento.com/blog/build-on-momento-tips-and-tricks-for-a-lightning-fast-game-leaderboard), a [sorted set](../../cache/develop/basics/datatypes#sorted-sets) is a collection data type that stores multiple value/score elements. In each element, the value is a `string` and the score is an `integer`. Specifically with instant messaging, you can store the time the message was sent as the *score* and a stringified JSON object as the *value*. Storing the message timestamp as the score will automatically allow you to sort messages by time since sorted sets do that by default.
+
+:::tip
+To store the time as the score, you must use an integer representation of the time. Store either ticks or time in milliseconds since the Unix epoch.
```
eaddingtonwhite: nit: maybe link out somewhere describing `milliseconds since the Unix epoch.` since guessing alot front end devs reading this they might not be as familiar?

docs/topics/patterns/instant-messaging.mdx
```diff
@@ -0,0 +1,355 @@
...
+
+Elements in a list do not need to be unique. You can push the same message into a list over and over again and it will be added every time. Lists are a good choice for instant messaging for the following use cases:
+
+* Paging messages N at a time
+* Adding multiple messages at once
+* Maintaining a chat history of fixed length, i.e. the last 100 messages
+
+When using a list, you have more array manipulation functions than with a sorted set, allowing you to automatically truncate message history if it gets too long or remove specific elements for things like content moderation.
+
+### Delivery
```
pratik151192, line 60: Topics will facilitate real-time communication, and cache will facilitate loading historical messages for new subscribers, right? Can we call this out explicitly somewhere? It is implicit, just took me 30-60 seconds to think through. So maybe it will help others if we are explicit.

pratik151192 (dismissed):
LGTM just few questions and nits

eaddingtonwhite (dismissed):

FractalJackal (dismissed):

docs/topics/patterns/instant-messaging.mdx
allenheltondev: Sending? In what context? That word does not seem intuitive to me.

docs/topics/patterns/instant-messaging.mdx
allenheltondev: I haven't heard anyone ask for it yet, but maybe they will when they read this?

cprice404:
the build failure is our fault, i will add a commit to fix it

docs/topics/patterns/instant-messaging.mdx
```diff
@@ -0,0 +1,359 @@
...
+Instant messaging is separated into two parts:
+
+* Message storage
+* Delivery
+
+These core components allow users to join your chat at any time and view the message history and enable them to receive new messages in real time.
+
+### Message storage
+
+Most instant messaging platforms store messages to allow users to view the conversation history if messages were sent prior to their arrival. Part of the startup tasks when entering a chat room is loading said history and displaying it on screen. This can easily be done with Momento in two ways: with a **sorted set** or a **list** cache item, each with their own tradeoffs.
```
cprice404: After reading over this section, I'm not sure that I understand:
* why someone would choose to use sorted sets over lists; lists seem easier and you don't really describe a downside that I can see.
* why we should mention two options here rather than just picking one and being prescriptive about it?

If we wanna keep both data types in here then I think maybe you could stand to put a little more info about when to choose one over the other. but tbh I think if your focus is to teach people how to get up and running with instant messaging as quickly as possible then it might be simplest just to cover it with lists?

cprice404: totally possible that I am just missing something here though!

docs/topics/patterns/instant-messaging.mdx
```diff
@@ -0,0 +1,359 @@
...
+
+* Paging messages N at a time
+* Adding multiple messages at once
+* Maintaining a chat history of fixed length, i.e. the last 100 messages
+
+When using a list, you have more array manipulation functions than with a sorted set, allowing you to automatically truncate message history if it gets too long or remove specific elements for things like content moderation.
+
+### Delivery
+
+Technically, message storage is not a necessary component for instant messaging. In theory you could have a messaging system that only showed the messages that were sent while a user is connected. Messages could be stored in memory in the client and you'd be done. That said, message delivery is a mandatory part of instant messaging. Getting the message from the sender to all recipients with minimal latency is the core of this pattern. To do this, we use [Momento Topics](https://www.gomomento.com/platform/topics).
```
cprice404: I might `s/That said/In contrast`

docs/topics/patterns/instant-messaging.mdx
```diff
@@ -0,0 +1,359 @@
...
+
+Momento Topics is a low-latency pub/sub service that connects clients together, clients to servers, servers to clients, and servers to servers. Think WebSockets but *way better*. When you publish a message to a topic, Momento will deliver it to all subscribers of the topic, providing fan-out capability or 1 to 1, depending on your use case.
```
cprice404: might be nice to get the word `broadcast` in here? maybe `Momento will broadcast`?

docs/topics/patterns/instant-messaging.mdx
```diff
@@ -0,0 +1,359 @@
...
+
+For instant messaging, the sender will *publish* a message to a topic. The recipients of the message, be it one or many depending on your use case, will be *subscribers*. Momento Topics will deliver the published message to the subscribers in real-time, providing for the instant messaging experience we're looking for!
+
+:::tip
+Topics are dynamic and are not defined resources in your infrastructure. The recommended way to build instant messaging is to *use the chat room or session identifier as the topic name*. This provides flexible and dedicated topics for chat use cases.
```
cprice404:
```suggestion
Momento Topics are dynamic and are not defined resources in your infrastructure. The recommended way to build instant messaging is to *use the chat room or session identifier as the topic name*. This provides flexible and dedicated topics for chat use cases.
```

docs/topics/patterns/instant-messaging.mdx
```diff
@@ -0,0 +1,359 @@
...
+  ```java
+  final TopicPublishResponse response = topics.publish("message-namespace", sessionId, "Hello world!").join();
+  ```
+  </TabItem>
+  <TabItem value="dotnet" label=".NET">
+
+  ```csharp
+  var response = await topics.PublishAsync("message-namespace", sessionId, "Hello world!");
+  ```
+  </TabItem>
```
cprice404, line 110: this is a nit and not a blocker for this PR, but it'd be great to define an example snippet for this, like what we do elsewhere in the docs. (e.g. https://docs.momentohq.com/topics/develop/api-reference#publish ). It would help us make sure that the code actually compiles and help us keep track of where we are missing examples for given languages. But I realize that it is a bit of extra work to go put the snippets into the SDK repos.

docs/topics/patterns/instant-messaging.mdx
```diff
@@ -0,0 +1,359 @@
...
+  });
+  ```
+  </TabItem>
+</Tabs>
+
+## Security
+
+Momento provides a [robust authorization mechanism](../develop/authentication/tokens) that limits consumers to specific resources and actions. When building an instant messaging feature, it is incredibly important to take scope into consideration so users do not intercept messages they shouldn't have access to.
+
+The pattern described above uses both cache and topic resources. Following the best practice recommended above by making our chat room the topic name, we can create policies in a couple of different ways.
```
cprice404: "making our chat room the topic name" - that reads a little confusing to me. "making our topic name match the name of our chat room"? or something like that?

docs/topics/patterns/instant-messaging.mdx
```diff
@@ -0,0 +1,359 @@
...
+      }
+    },
+    {
+      "role": "publishsubscribe",
+      "cache": "instant-messaging",
+      "topic": "{chatid}"
+    }
+  ]
+}
+```
```
cprice404: what is this? I don't think we actually support any JSON representation of policies, and this section of the doc doesn't really explain to a user how they would achieve this policy. Maybe that's okay if you are just trying to convey the theoretical concepts but I don't think a user could take this and get an app working from it, don't know if that is part of the goal here?

docs/topics/patterns/instant-messaging.mdx
```diff
@@ -0,0 +1,359 @@
...
+      "role": "read",
+      "cache": "instant-messaging",
+      "item": {
+        "key": "{chatid}"
+      }
+    },
+    {
+      "role": "publishonly",
+      "cache": "instant-messaging",
+      "topic": "{chatid}-raw"
```
cprice404: raw is a reasonable word here, but wondering if it might be a little more clear if you said `unmoderated` or `input` or something.

cprice404 (commented):
left a handful of comments. i'm not entirely in the loop on your goals for this page so take them or leave them as you see fit!

docs/topics/patterns/instant-messaging.mdx
allenheltondev: Yeah that's a great point, i wanted to be complete but as you mention, this is for guidance. i'll take out sorted sets

docs/topics/patterns/instant-messaging.mdx
allenheltondev: I'm not a fan of this suggestion because it mixes the product name "Momento Topics" with "topics" the...thing in the product. I will reword for less confusion though

eaddingtonwhite (approved):

PR #909, opened 2024-04-24
feat: add momento-drupal integration to drupal page

rishtigupta:
## PR Description:
This commit 
- Adds momento-drupal integration link to drupal integration page. 
- There were a few reorganizes in the rust sdk for topic client ([ref](https://github.com/momentohq/client-sdk-rust/tree/main/src/topics)). This commit also corrects the code to correctly pull the topic client for rust sdk such that the build succeeds.

## Issue
https://github.com/momentohq/dev-eco-issue-tracker/issues/790

### DIFF
--- a/docs/cache/develop/integrations/drupal-integration.md
+++ b/docs/cache/develop/integrations/drupal-integration.md
@@ -7,6 +7,16 @@ pagination_next: null
 ---
 
 # Integrating Momento Cache with Drupal
+# METHOD 1: (Recommended)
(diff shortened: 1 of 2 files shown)


docs/cache/develop/integrations/drupal-integration.md
```diff
@@ -7,6 +7,16 @@ pagination_next: null
 ---
 
 # Integrating Momento Cache with Drupal
+# METHOD 1: (Recommended)
```
pgautier404: We should completely deprecate/disappear the module hosted in our repository, so there should only be one method for integrating with Drupal. I think it would make more sense to just remove the old content from this file. It looks like everything in here is in the README you linked to so you shouldn't need to preserve any of it.

cprice404 (approved):

PR #926, opened 2024-04-30
feat: add a compression page for the Node.js SDK

nand4011:
Add a page showing how to enable and use compression in the Node.js SDK.

Add links to the compression page in the main Node.js SDK page and the SDK overview page.


cprice404 (commented):
this is great, thanks! let's hold off on merging it while the dust settles on the zstd vs zlib conversation, but it should be only minor changes if we do need to tweak it. thank you!

rishtigupta (approved):

PR #997, opened 2024-05-17
chore: remove setIfNotExists now that we have setIfAbsent and friends

cprice404:


malandis (approved):
Would be useful to have a linked task to remove this from all the SDKs that had it.

cprice404:
> Would be useful to have a linked task to remove this from all the SDKs that had it.

I'm not necessarily planning on removing it from the SDKs since that would be a breaking change. Marking it as deprecated in favor of SetIfAbsent seems sufficient for the forseeable future. I think removing from the docs is okay though.

PR #1017, opened 2024-05-28
chore: update CAS section with warning

pratik151192:


docs/cache/develop/api-reference/index.mdx
```diff
@@ -261,6 +261,11 @@ See [response objects](./response-objects.md) for specific information.
 
 ### SetIfAbsent
 
+:::warning
+Do NOT use compare-and-set (CAS) APIs such as `SetIfAbsent` with scalar or non-CAS APIs such as `Set`. Doing so will result
```
krispraws: Can we call out `Delete` also as a non-CAS API here?

krispraws (commented):
We are missing the warning for SetIfEqual and SetIfNotEqual.

krispraws (dismissed):
Nitpicky - I have heard of Redis check-and-set, and compare-and-swap. Is compare-and-set widely used?

pratik151192:
@krispraws eww you're right; CAS traditionally in my mind comes from CPU provided CAS instruction which is compare-and-swap. updated

krispraws:
> @krispraws eww you're right; CAS traditionally in my mind comes from CPU provided CAS instruction which is compare-and-swap. updated

Aren't we closer to the Redis terminology though? `swap` to me implies atomicity, which we cannot guarantee.

pratik151192:
@krispraws  TIL Redis `check-and-set`.. updated!

krispraws (approved):

PR #1026, opened 2024-06-14
add limits index page

tvald:
Adds a limits index page so that we can share a single link with customers who want to learn about limits, rather than a different link for each product.


tvald (commented):
use ReferenceCard instead of Card

anitarua (commented):
lgtm overall! should there be a nav bar link as well for limits to match the sidebar though?
Screenshot 2024-06-14 at 9 16 10 AM

tvald:
> lgtm overall! should there be a nav bar link as well for limits to match the sidebar though?

The top nav is getting a bit cluttered, so I'll need to revisit the information architecture as we add in Storage. For now, we just need this page so that sales / support can send a single link to folks who ask about service limits.

anitarua (dismissed):

anitarua (approved):

PR #1029, opened 2024-06-14
Upgrade packages to latest versions.

tvald:
This is a fairly large upgrade, though for the most part it's just fixing problems uncovered by stricter parsing rules and better validation in the latest docusaurus.

Project changes:
- upgrade all packages to latest 
- switch to TS docusaurus config
- rewrite some of the code injection with stricter validation and type safety

Content changes:
- fix a bunch of broken links
- fix some issues that were probably previously broken and silently ignored, such as unescaped special characters or unclosed JSX tags


tvald:
I spot-checked a bunch of pages, doing side-by-side comparisons with existing docs. I have checked all code injection component types, as well as more complex pages like the home page, top-level products pages, and the SDK list.

docs/sdks/dart/topics.mdx
```diff
@@ -87,5 +87,5 @@ This is an example of subscribing to a topic called "topic", then catching the r
 You can find complete, working examples in the [Dart SDK GitHub repo examples directory](https://github.com/momentohq/client-sdk-dart/tree/main/example).
 
 :::info
-Beyond these basic API calls check out the [API reference page](/develop/api-reference) for more information on the full assortment of Momento API calls.
+Beyond these basic API calls check out the [API reference page](/topics/develop/api-reference) for more information on the full assortment of Momento API calls.
```
cprice404, line 90: hrm, I could have sworn docusaurus was supposed to catch these. thanks for fixing.

plugins/example-code-snippets/src/inject-example-code-snippets.ts
```diff
@@ -1,140 +1,138 @@
...
-    return;
+    throw Error('Missing attribute SdkExampleCodeBlock.language');
   }
   const language = exampleLanguage(snippetLanguageAttr);
 
-  const snippetId = value.match(/snippetId={'([^']+)'}/m)?.[1];
-  const filename = value.match(/file={'([^']+)'}/m)?.[1];
-  if (snippetId === undefined && filename === undefined) {
-    console.log(
-      `Unable to find "snippetId={'xxx'}" or "file={'xxx'}" attribute on jsx tag:\n${value}`
```
cprice404, line 121: These log statements were intended to give us a way to audit all the code blocks in the docs and swarm to fill in missing ones, someday, when time permits. Did you have another idea in mind for how we can easily identify the list of example snippets that are missing for a given language?

cprice404 (commented):
Thanks for upgrading the deps, and finding / fixing the broken links! much appreciated.

I'm +1 on this except for the question about how to identify a list of missing examples that we need to go add in a given language.

plugins/example-code-snippets/src/inject-example-code-snippets.ts
tvald: This is actually the error produced when an `SdkExample*` element is being parsed. In this specific case, it's complaining that the element is missing required attributes (either `snippetId` or `file`).

Missing snippet / file detection is still handled exactly the same way, and `yarn clean && yarn build` will dump out a report.

I generally converted `console.log()` to `throw Error()` in the code injection plugin so that we're not quietly skipping any misconfigured `SdkExample*` elements.

docs/sdks/dart/topics.mdx
tvald: The latest parser is stricter about links and by default doesn't permit linking to a redirect. We have a redirect for `/develop/api-reference -> /cache/develop/api-reference`, so even though the link wasn't technically broken here, it was wrong. 😓

PR #1063, opened 2024-07-12
Chore/move integration directory

riokwmt:
Moved the integrations directory in the JP directory out to under /cache.


poppoerika:
`integrations` directory has already moved under the `cache`; therefore no file change was made.

PR #1104, opened 2024-07-18
update api matrix for Go auth client, replace TokenScope with PermissionScope

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/954

Adds the Go sdk auth client file to the api support matrix parser. GenerateDisposableToken, GenerateApiKey, and RefreshApiKey should all appear once this is merged.

Also replaces mentions of [TokenScope (deprecated)](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/auth/tokens/permission-scope.ts#L130-L133) with PermissionScope

PR #1111, opened 2024-07-22
Update auth.md

yushimomentohq:
update api matrix for Go auth client, replace TokenScope with PermissionScope #1104


yushimomentohq:
It was overlooked that the following was not a merge but an Open state.
I shall do this again at a later date, once Close.
https://github.com/momentohq/public-dev-docs/pull/1104

PR #1116, opened 2024-07-22
Update index.md

yushimomentohq:
chore: add rust sdk storage examples #1090


yushimomentohq:
It was overlooked that the following was not a merge but an Open state.
I shall do this again at a later date, once Close.
https://github.com/momentohq/public-dev-docs/pull/1090

PR #1130, opened 2024-08-05
chore: delete duplicated language support page

poppoerika:
I see Language Support has its own directory yet the single page existed under the API Reference directory.
So I'm deleting the one under that.


cprice404:
thx @poppoerika , i think this is the right thing to do but there are a few links that need to be changed to `../`. I can try to fix them soonish if I get to them before you do.

cprice404 (approved):

PR #1132, opened 2024-08-06
Fix typo in Japanese document

Lucky3028:


Lucky3028:
Could you teach me why this PR is ignored?

tvald (approved):

PR #1142, opened 2024-08-12
chore: Add the cloud linter optional services

nand4011:
Add the Amazon services that cloud linter can optionally check, as well as calls it makes.


docs/cloud-linter/index.md
```diff
@@ -56,6 +60,14 @@ DynamoDB:
 CloudWatch:
 - GetMetricData: 20 TPS
 
+S3:
+- ListBuckets
+- ListBucketMetricsConfigurations: 10 TPS
+
+API Gateway:
+- GetRestApis: 10 TPS
+- GetRestApi
```
nand4011: @pgautier404 I didn't see a limit on this call in the cloud linter code. Do you remember if that is done in a different way?

pgautier404: @nand4011 The methods for processing those resources are using the metrics and control plane limiters in use throughout. Let me know if that doesn't address the question and we can take a look at it.

cprice404 (commented):
do we need to update the diagram at all? or are you thinking we can skip that because the diagram now reflects the default behavior?

nand4011:
I updated the diagram.

docs/cloud-linter/index.md
```diff
@@ -42,6 +42,10 @@ Cloud Linter is intended to interact with the following AWS services via their r
 - Amazon ElastiCache
 - Amazon CloudWatch
 
+It can optionally interact with the following AWS services:
+- Amazon S3, using the `-enable-s3` option
+- Amazon API Gateway, using the `-enable-api-gateway` option
```
pgautier404:
```suggestion
- Amazon S3, using the `--enable-s3` option
- Amazon API Gateway, using the `--enable-api-gateway` option
```

cprice404 (approved):

PR #1165, opened 2024-08-23
fix: update topics webhook limits

anp13:


danielamiao:
@tvald @cprice404 FYI during a recent review of our limits @anp13 discovered the docs do not reflect what our service is enforcing. I checked webhook metrics and no customer is regularly sending more than 5 TPS of webhook deliveries, `bot@dipway.com` does occasionally exceed it but getting throttled since our actual limit is 5 TPS.

As a result, we should just update the docs to reflect reality. Let me know if you have any concerns

cprice404 (approved):
ok

PR #1300, opened 2024-09-03
Beginning section for data-driven API section in the docs. 

allenheltondev:
Built an example and a builder for docs from a json schema
The API section is intended to incorporate API definitions across all products, similar to how Stripe approaches the API reference. We will still have product focused sections in the docs, but that's more driven toward use cases, workflows, and quick starts (all future)
I also have in there an example of how to link to sections inside of a page from the navbar :fire:

**NOTE: This is not completely done, but I'm moving off of this for now. The new content is not linked from the primary docs definition, so it's undiscoverable by obscurity**

If you want to pull it down and look, you can go to http://localhost:3000/api


cprice404 (commented):
i will pull this down and play with it sometime soonish when time permits, sounds cool!

the build is failing, haven't looked into why yet. once that is resolved i have no problem with shipping this dark.

eaddingtonwhite (approved):

PR #1309, opened 2024-09-09
Subscribe HTTP docs and Topic page move

allenheltondev:
Updated the HTTP API docs to include Subscribe
Moved the API Reference directory to the root
Moved the Metrics page to the root instead of the only thing under "Manage"
Updated a little bit of the verbiage on the metrics page.

*Note* I updated the japanese pages because it wouldn't build if i didn't. I don't know if that's something I was supposed to do or not.


docs/topics/api-reference/http-api.md
```diff
@@ -45,75 +56,179 @@ Publishes a message to a topic.
...
+      "item": {
+        "topic_sequence_number": 0,
+        "value": {
+          "text": "hello world"
+        }
+      }
+    }
+  ]
+}
+```
```
eaddingtonwhite, line 232: Just a nit but maybe put the simpler example w/ no sequence number specified first? Feels bit odd to put the simpler example below a more complicated example passing sequence number and getting discontinuity back?

eaddingtonwhite (dismissed):
Looks good. Just small nit on topic REST example ordering. Fine fixing in follow up PR.

eaddingtonwhite (approved):

PR #1320, opened 2024-09-20
chore: remove remaining occurrences of `instanceof` from JS examples

cprice404:
This commit replaces a bunch of hard-coded/inlined JS example snippets
with references to the examples in the repository. The repository
versions are updated to use the new `switch` pattern rather than the
old `if/instanceof` pattern.


rishtigupta (approved):
:shipit:

malandis (approved):

PR #1349, opened 2024-10-08
Add mediastore origin page

allenheltondev:
Adding page to define Origins in Mediastore docs

http://localhost:3000/mediastore/core-concepts/origin


eaddingtonwhite (approved):

allenheltondev:
If you'd rather see the rendered version
image

PR #1437, opened 2024-12-17
fix: adjust rust source locations for repo refactor

malandis:
To fix the release plugin, in the Rust SDK repo we lifted the SDK
source from `./sdk` to the repo root. We adjust the source paths here
to fix the docs.


eaddingtonwhite (approved):

tvald (changes requested):
land #1435 first

PR #1444, opened 2024-12-18
chore: add tokenId optional parameter for generateDisposableToken docs

anitarua:
Will be pointed to by the rust topics example updated to show `publisher_id`: https://github.com/momentohq/client-sdk-rust/pull/418


pgautier404 (dismissed):

nand4011 (dismissed):
My only nit is that the capitalization, presence of a starting article, and punctuation are different between the parameters, optional parameters, and return values.

rishtigupta (dismissed):
:shipit:

anitarua:
> My only nit is that the capitalization, presence of a starting article, and punctuation are different between the parameters, optional parameters, and return values.

I think I got them all, but let me know if you see more!

nand4011 (dismissed):

nand4011 (approved):
Looks good!

PR #1470, opened 2025-01-17
chore: add Java to the leaderboards API reference

nand4011:
Add Java to the leaderboards API reference.

Remove unneeded auth file from the leaderboards docs.


anitarua (approved):
🎉

PR #1479, opened 2025-03-05
[Preview] Functions

allenheltondev:
**Please do not translate these pages! These are for reference only.**

Based on the infromation that has been tossed around, I came up with a set of pages that describe *Functions* and how users can interact with them. I made lots of assumptions and would love to hear feedback on what is incorrect, up in the air, or what was missed. 

The docs are written from a perspective that *Functions* have already launched and we have a first-pass experience with them.

If you run the docs locally, you can view the root at `http://localhost:3000/functions`


docs/functions/authentication/authentication-methods.mdx
```diff
@@ -0,0 +1,82 @@
...
+## Session tokens
+Session tokens are short-lived credentials that provide temporary access to *Momento Functions*. They are useful for time-limited operations, user sessions, or scenarios where API keys should not be exposed, like interactions in a user interface.
+
+Tokens are automatically expired based on their configured lifespan, with a *max valid duration of 1 hour*. These tokens cannot be renewed, meaning a new token must be issued when the previous one expires.
+
+All session tokens are fine-grained access control (FGAC) tokens, limited only to data-plane operations. They cannot perform actions like creating or deleting a function, only invocation.
+
+## Usage
+
+Both API keys and session tokens must be provided in every request made to a function and are valid in one of two places:
```
kvcache, line 32: you don't provide both an api key and a session token.

docs/functions/authentication/authentication-methods.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+### Invoke all functions in a namespace
+The following example grants access to invoke all functions within the namespace `"my-namespace"`.
+
+```JSON
+{
+  "permissions": [
+    {
+      "role": "invoke",
+      "namespace": "my-namespace",
```
kvcache, line 57: don't we use the word "cache" still?

docs/functions/authentication/authentication-methods.mdx
```diff
@@ -0,0 +1,82 @@
...
+      "functions": ["*"]
+    }
+  ]
+}
+```
+
+### Invoke specific functions in a namespace
+To grant access to invoke only `"function-a"` and `"function-b"` within the namespace `"my-namespace"`, define the permissions as follows:
```
kvcache, line 65: This statement and the json below are inconsistent with one another

docs/functions/authentication/aws-role-delegation.mdx
```diff
@@ -0,0 +1,137 @@
...
+  - permissions
+---
+
+# Configuring AWS role delegation for Momento Functions
+To allow *Momento Functions* to access Amazon S3 or Amazon Bedrock, you need to configure a trust relationship between your AWS account and Momento. This setup enables Momento to assume an IAM role in your AWS account with only the permissions you define.
+
+Momento supports multiple IAM roles, allowing you to configure separate roles for different use cases, services, or functions.
+
+## How AWS role delegation works
+To enable *Momento Functions* to access your AWS services, you must create an IAM role in your AWS account that grants the necessary permissions. This role must include a trust policy that allows Momento to assume the role when needed. Once the role is created, you link it to your Momento account by pasting the IAM Role ARN into the Momento onsole. After the role is successfully configured, Momento can securely access AWS services according to the permissions specified in the IAM policy.
```
kvcache, line 23: This whole page is specific, but I don't know if it is accurate.

docs/functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+  - functions
+  - wasm
+  - real-time processing
+  - event-driven
+  - serverless
+---
+
+# Momento Functions
+
+*Momento Functions* are a high-speed compute layer designed for processing massive amounts of real-time data. Whether you're aggregating IoT sensor data, handling financial transactions, or enriching event streams with AI, *Momento Functions* allow you to process, transform, and act on data instantly—without managing infrastructure.
```
kvcache, line 19:
```suggestion
*Momento Functions* is a lightweight compute layer designed for processing real-time data. Whether you're aggregating IoT sensor data, recording financial transactions, or enriching event streams with AI, *Momento Functions* allows you to process, transform, and act on data instantly—without managing infrastructure.
```
Looks good, but I'd tighten this up a little further.

docs/functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+Built for extreme scale and performance, functions offer low-latency execution and shared global memory across all running function instances. Unlike general-purpose serverless compute, they do not execute in isolation, allowing all instances of a function to share state for efficient data processing.
```
kvcache, line 21: I think this wording isn't quite right for this implementation detail, and I'm struggling to come up with an approach that succinctly captures the differentiation's semantics. It might be best to either pull this paragraph or iterate on it more.

docs/functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+```mermaid
+mindmap
+  root((Momento Functions))
+    (Fast Execution)
+      (5ms SLA for ultra-fast processing)
```
kvcache, line 27: I don't think we're offering an SLA. Some functions will take much longer to execute than 5ms, and it's largely outside of Momento's control how long they take. We can't make Bedrock go any faster than it does, and we can't exceed the universe's rate of information propagation (C).
Maybe we can say something about sub-millisecond cold starts, if you want to add a latency number? I think people typically associate that number with the right technical detail.

docs/functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+      (10x faster, 10x cheaper)
```
kvcache, line 28: Than...? 😄

docs/functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+    (Shared Global Memory)
+      (All function instances share state)
```
kvcache, line 30: This is sort of the case, and I'm not sure how far we'll go down the path of pretending that an in-memory cache is global. We could make it global and transactional, but that would slow down functions tremendously!

docs/functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+      (Persistent between executions)
+    (Batch Processing)
+      (Dynamically process batched events)
+      (Seamless data aggregation)
+    (Seamless Integrations)
+      (Amazon S3)
+      (Momento Topics)
+      (Momento Cache)
+      (Amazon Bedrock)
```
kvcache, line 39: Maybe you want to either alphabetize or group the integrations by top-level owner?

docs/functions/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+```
+
+## What makes Momento Functions different?
+Traditional serverless platforms execute functions in isolation, requiring external coordination for batching, state sharing, and data transformations. Momento Functions remove these limitations:
+
+* Global memory shared across all function instances enables stateful execution without an external database.
+* Automatic batching allows functions to process multiple related events in a single execution cycle.
+* No concurrency limits means functions can instantly scale to handle millions of invocations per second without cold starts.
```
kvcache, line 47: I'm not sure we'll have no concurrency limits. I think it's going to be easy to set high limits, but we can't sustain, e.g., a trillion concurrent requests on each frontend host.

kvcache (commented):
I got partway through, but I need to go run an errand. We might should iterate on this further internally rather than on the public repo. Thanks for starting this!

PR #1486, opened 2025-04-04
Proxy updates

allenheltondev:
Adding additional verbiage around memcached and a small tutorial for running as a sidecar

also fixed some broken links


docs/cache/proxy/configuration/sidecar.md
```diff
@@ -0,0 +1,101 @@
...
+import java.net.InetSocketAddress;
+
+public class App {
+    public static void main(String[] args) throws Exception {
+        // If running in Kubernetes, use localhost. In Docker Compose, use "proxy".
+        MemcachedClient client = new MemcachedClient(
+            new InetSocketAddress("localhost", 11211)
+        );
+
+        client.set("foo", 3600, "bar");
```
krispraws: This returns a [future](https://javadoc.io/doc/net.spy/spymemcached/latest/net/spy/memcached/MemcachedClient.html#set-java.lang.String-int-java.lang.Object-). You have to call `.get()` on it.

eaddingtonwhite (dismissed):

docs/topics/api-reference/auth.md
```diff
@@ -302,7 +302,7 @@ const tokenResponse = await authClient.generateDisposableToken(
 <details>
     <summary>Do these tokens control access to the Momento control plane APIs?</summary>
 
-    Access tokens generated with the [GenerateApiKey](#generateapikey-api) API only control access to the Momento data plane APIs. A token for access to Momento's control plane APIs must be generated using the [Momento console](https://console.gomomento.com/).
+    Access tokens generated with the [GenerateApiKey](#generateapikey) API only control access to the Momento data plane APIs. A token for access to Momento's control plane APIs must be generated using the [Momento console](https://console.gomomento.com/).
```
krispraws: Did this change? I do not remember it being like that. Superuser tokens generated via GenerateApiKey used to work for control plane APIs. Only disposable tokens are retricted to data plane.

allenheltondev: No, looks like that paragraph is wrong. i'll update that real quick

docs/cache/proxy/configuration/sidecar.md
allenheltondev: Just to make sure I'm doing it right, like this @krispraws ?
`client.set("foo", 3600, "bar").get();`

krispraws: Yes

eaddingtonwhite (approved):

PR #1488, opened 2025-04-07
chore: set if hash methods accept hash value as bytes

anitarua:
minor fix after realizing the comparison hash values should be accepted as bytes


rishtigupta (approved):
:shipit:

PR #1497, opened 2025-11-19
Adding RESP compatibility page

allenheltondev:
Added the RESP page.

I also removed the old "introduction" pages that were stale and duplicate of newer content. Redirects were added for these pages as well.


danielamiao:
Do we publish our cell endpoints somewhere? We should with this update so people know which endpoints to use with which regions.

allenheltondev:
@danielamiao I updated the region page to include the cell name instead of api endpoint then i made a new section beneath to show how to use the cell name for the API and for RESP 👍

danielamiao:
> @danielamiao I updated the region page to include the cell name instead of api endpoint then i made a new section beneath to show how to use the cell name for the API and for RESP 👍

Awesome thank you, a couple more notes:
1. we have a new region added as of last month `cell-1-eu-north-1-1`
2. you should put "RESP and SDK endpoints". We will still support using our SDK to connect via grpc or protosocket-rpc.
3. idk if this is worth calling out but we support RESP2 only today, support for RESP3 is on the roadmap, contact us to learn more.

eaddingtonwhite (approved):

PR #1501, opened 2025-12-17
Update gif on docs for getting started

allenheltondev:
Swap from a video to a gif showing the new console for getting started with cache

I took out the api key creation portion of the gif because that part was not relevant here. It's now just getting/setting items in the cache via the console as well as creating a cache


danielamiao (approved):
thank you, i like this better!

PR #1507, opened 2026-02-07
chore: add momento cluster docs

danielamiao:
tested create, list, delete.

### DIFF
--- a/docs/valkey/http-api.md
+++ b/docs/valkey/http-api.md
@@ -54,7 +54,7 @@ Creates a new Valkey Cluster or updates an existing one with the specified confi
 ```json
 {
   "description": "My production cluster",
-  "node_instance_type": "m6g.large",
+  "node_instance_type": "cache.m6g.large",
   "shard_count": 3,
   "replication_factor": 2,
   "enforce_shard_multi_az": true,
@@ -90,7 +90,7 @@ Creates a new Valkey Cluster or updates an existing one with the specified confi
 {
   "name": "my-cluster",
   "description": "My production cluster",
-  "node_instance_type": "m6g.large",
+  "node_instance_type": "cache.m6g.large",
   "shard_count": 3,
   "replication_factor": 2,
   "enforce_shard_multi_az": true,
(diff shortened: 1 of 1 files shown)


docs/valkey/http-api.md
```diff
@@ -0,0 +1,344 @@
...
+| Field | Type | Description |
+|-------|------|-------------|
+| name | String | The name of the cluster. |
+| description | String | The cluster description. |
+| node_instance_type | String | The instance type for cluster nodes. |
+| shard_count | Integer | The number of shards. |
+| replication_factor | Integer | The number of replicas per shard. |
+| enforce_shard_multi_az | Boolean | Whether multi-AZ placement is enforced. |
+| shard_placements | Array | The placement configuration for each shard. |
+| status | String | The cluster status: `Creating`, `Active`, or `Deleting`. |
```
ben-kugler: We should add 'Updating' here as well, this will be needed when we support `Update`

https://github.com/momentohq/mr2rs/pull/2828

docs/valkey/http-api.md
```diff
@@ -0,0 +1,344 @@
...
+
+- The cluster was successfully deleted.
+
+#### Error
+
+*Status Code: 401 Unauthorized*
+- This error type typically indicates that the Momento API key passed in is either invalid or expired.
+
+*Status Code: 404 Not Found*
+- The specified cluster does not exist.
```
ben-kugler: delete will never return 404, even if cluster isn't found

docs/valkey/http-api.md
```diff
@@ -0,0 +1,292 @@
...
+#### Success
+
+*Status Code: 200 OK*
+
+```json
+[
+  {
+    "name": "my-cluster",
+    "description": "My production cluster",
+    "node_instance_type": "m6g.large",
```
ben-kugler: `cache.m2g.large`, etc?

Unless we want to append `cache.` for node types if it doesn't have it? Either way, current error type will show valid instances with `cache.`

danielamiao: good call i forgot this update

ben-kugler (approved):

PR #1509, opened 2026-03-06
chore: update cluster docs

danielamiao:


docs/valkey/http-api.md
```diff
@@ -25,35 +25,29 @@ The API Key must be provided in the `Authorization` header.
...
+|------------------|-----------|--------|-----------------------------------------------------------------------------------------------------|
 | Authorization    | yes       | String | The Momento API key, in string format, is used for authentication/authorization of the request.    |
 | Content-Type     | yes       | String | Must be `application/json`.                                                                        |
 
 #### Request Body
 
 ```json
 {
-  "description": "My production cluster",
+  "name": "my-cluster",
```
rlinehan, line 50: Not introduced in this pr, but this example json is not actually valid, because there aren't 3 shards. If you change the `shard_count` to 1 it would be, or add 2 more `shard_placements`

docs/valkey/http-api.md
```diff
@@ -101,22 +94,19 @@ Creates a new Valkey Cluster or updates an existing one with the specified confi
       "replica_availability_zones": ["us-east-1b", "us-east-1c"]
     }
   ],
-  "status": "Creating",
-  "errors": []
```
rlinehan, line 105: Similar comment here, this isn't actually a valid configuration.

docs/valkey/http-api.md
```diff
@@ -132,25 +122,135 @@ Creates a new Valkey Cluster or updates an existing one with the specified confi
...
+
+#### Success
+
+*Status Code: 200 OK*
+
+```json
+{
+  "name": "my-cluster",
+  "node_instance_type": "cache.m6g.large",
+  "shard_count": 3,
```
rlinehan, line 162:
```suggestion
  "shard_count": 1,
```

docs/valkey/http-api.md
danielamiao: i added 2 more to shard placements to have a more illustrative example

docs/valkey/http-api.md
danielamiao: updated

docs/valkey/http-api.md
```diff
@@ -99,24 +102,31 @@ Creates a new Valkey Cluster or updates an existing one with the specified confi
...
+    {
+      "shard_index": 2,
+      "availability_zone": "us-east-1a",
+      "replica_availability_zones": ["us-east-1b", "us-east-1c"]
     }
   ],
-  "status": "Creating",
-  "errors": []
+  "status": "Creating"
 }
```
anp13, line 102: This is not what we will see in response for Creating. This is a sample response for Creating now:
```
{
  "name": "test-ec-clusters",
  "enforce_shard_multi_az": true,
  "shard_count": 0,
  "replication_factor": 0,
  "shard_placements": [],
  "node_instance_type": "cache.m6g.xlarge",
  "status": "Creating"
}
```

I think we should not return any response for Create because it is weird that the fields are wiped out in the response. But today we return this.

docs/valkey/http-api.md
```diff
@@ -265,28 +575,101 @@ curl -X PUT -H "Authorization: <token>" \
...
         "shard_index": 1,
-        "availability_zone": "us-east-1b",
-        "replica_availability_zones": ["us-east-1a", "us-east-1c"]
+        "availability_zone": "us-east-1a",
+        "replica_availability_zones": ["us-east-1b", "us-east-1c"]
       }
     ]
   }' \
-  "https://api.cache.cell-1-us-east-1-1.prod.a.momentohq.com/cluster/my-cluster"
+  "https://api.cache.cell-1-us-east-1-1.prod.a.momentohq.com/ec-cluster"
```
anp13, line 570: nit: May be we should update this example to have 3 shards since the update shard config example assumes that we are starting with 3 shards and going to 5.

anp13 (approved):

PR #1517, opened 2026-05-06
chore: functions docs

kvcache:
Here's my attempt to capture the essence of Functions for the dev
docs site.

PR #1518, opened 2026-05-07
chore: update functions docs

kirmar:
Thank you so much for #1517, @kvcache!! I didn't review it in time to suggest the CLI version of `invoke-function` (which I also need to add to https://github.com/momentohq/functions/pull/92), so just sharing here alongside a couple other changes.

I'm still getting `yarn` to work, so if someone already has it set up to make sure this is still formatted correctly, that would be appreciated :pray:


kvcache (approved):
I have validated that this looks right on my laptop.

PR #1525, opened 2026-07-01
fix: klog sample 0 does not disable klog

anitarua:


anitarua:
already fixed in a later version of pelikan, this update no longer needed

